PR #44, opened 2014-12-30
fix(rest-api): The names of query parameters are fixed.

jvaldesh:
In REST API: Get Definition count
The query parameters version and latestVersion are invalid. The valid parameters are ver and latest respectively.


meyerdan:
Thanks!

Issue for tracking: https://app.camunda.com/jira/browse/CAM-3265

PR #57, opened 2015-09-09
Using historyService instead of runtimeService

facundofarias:
When I was reading the docs, it was a bit confusing, so because of that here is the pull request that fixes this documentation issue (also because the method doesn't exist on runtimeService but it does on historyService).


meyerdan:
Thanks for the fix :)

facundofarias:
No problem! :+1:

PR #59, opened 2015-10-19
Describe process instance modification at start

ingorichtsmeier:
show the possibility to start a process instance at a certain point.


meyerdan:
Great, thanks Ingo

PR #64, opened 2015-11-10
chore(variables) how to change a file value

ingorichtsmeier:
short clarification answer a recent question on the forum


meyerdan:
Thanks ingo.

I added a headline for your content. This way it is easier for users find it if the search for "Changing a File Value"

PR #66, opened 2015-12-21
wrong API used

ingorichtsmeier:
Added the missing decision ID to the API call


saig0:
Hi Ingo,

thank's for reporting. Can you change the "decisionID" to "decisionKey" because it is called so in the JavaDocs?

ingorichtsmeier:
Hi Philipp, now you can merge the change and delete the branch afterwards.

saig0:
Thank you!

PR #77, opened 2016-04-01
changed varId to varName

berndruecker:
to reflect what the method is really doing


ThorbenLindhauer:
Hi Bernd,

Thanks for the pull request. There are various other places where `{varId}` is used although `{varName}` would be more appropriate (e.g. https://docs.camunda.org/manual/latest/reference/rest/execution/local-variables/put-local-variable-binary/). Are you willing to update these, too?

Cheers,
Thorben

ThorbenLindhauer:
Merged: https://github.com/camunda/camunda-docs-manual/commit/afd098ec33a52eb150b62fee3b9dbcd1e6dcf23b

Thanks to both of you, @berndruecker and @mburke84

PR #81, opened 2016-06-09
fix(external-task): make description match code example

falko:
Code says 60*10000 which is 10 minutes.


menski:
Hi Falko,

thanks for spotting this error. I merged your fix with 7fdab133f0b55fd5b6f77ddbca16de9334c1ee5b. And also split up the time calculation to be more readable 403f5e0af5abd5f500523ac5d02563b6572f2e46.

Cheers,
Sebastian

PR #96, opened 2016-09-21
improve(embedded-forms): use Uri instead of $scope

falko:
$scope does not have to be injected as described above. Instead Uri is frequently used together with $http


SebastianStamm:
Thanks for the Pull Request 👍

PR #109, opened 2016-12-09
Wrap long code line

falko:
Wrap long code line so that the important part is visible. Otherwhise, the code is always hidden regardless of the Browser zoom.


mburke84:
Hi Falko,

Thanks for raising this pull request.

I have cherry-picked to commit to master, see:
https://github.com/camunda/camunda-docs-manual/commit/7db4e3cf4f516f5349697892005224db9da255e5

Thank you for your contribution and best regards,
Mat

PR #112, opened 2017-01-19
add hint to missing error boundary event

ingorichtsmeier:
From a discussion in the last training, I think this clarification may help other readers, too.


JoHeinem:
Hi Ingo,

Yes, I agree the hint makes sense. I added your pull request. 

Best,
Johannes

PR #121, opened 2017-03-27
Fix case definition typos

davidmichaelkarr:
Fixed a couple of typos in the reference pages for the rest api for case-definition.


davidmichaelkarr:
I'm confused. This has changes from my other branches/PRs.

PR #124, opened 2017-05-08
replace task.getTopic() with task.getTopicName()

FrVaBe:
The method getTopic() does not exist on LockedExternalTask objects.
I assume that getTopicName() is meant (as it is the only method
with a topic related result value) and changed the code samples
accordingly.


ThorbenLindhauer:
Thank you!

PR #129, opened 2017-06-01
added Loading scripts within other Scripts section

StephenOTT:


romansmirnov:
Hi @StephenOTT ,

Thanks for your contribution.

We decided not to merge your changes to the user guide, since you are documenting some functionality of Nashorn. We does not have such test cases, that verify that those kind of functionality works reliable. As a consequence, adding this to the user guide would mean that we have to support this.

If you are interested, you can write a blog post about that?

Cheers,
Roman

PR #133, opened 2017-06-30
The path is relative to the `app/tasklist/` folder

Tristan-ow:
It was incorrect to say that the path is relative to the app/tasklist/scripts folder as it is only relative to the  `app/tasklist/`


romansmirnov:
Thanks.

PR #136, opened 2017-07-18
Update delegation-code.md

parkerrob:
Added refined commentary re assignment listener and event lifecycle. In particular, unassigned can be determined by an assigned event where assignee is null, and an assignment event is only generated before a create event provided the assignee is not null.


tasso94:
Hi @parkerrob,

we discussed your contribution and concluded that we cannot merge your proposal. We must make sure that the documented functionality works reliable. Adding your contribution would mean that we have to support the functionality exactly as it is described.

Best,
Tassilo

PR #139, opened 2017-08-09
fix(databaseSchemaUpdate): describe actual behavior since version 6.0

falko:
IRC, automatic updates of the schema are disabled since version 6.0. We always write schemaVersion=fox into the ACT_GE_PROPERTY table. Therefore the engine cannot know if the database schema is in the correct version or needs to be updated. The schema version must be ensured by the admin.

I tried this, by running a 7.7.0-ee engine on a 7.6.1-ee schema with databaseSchemaUpdate=true. The engine started without errors, but failed later during the deployment of a process because columns added in 7.7 could not be found:
> ENGINE-03004 Exception while executing Database Operation 'INSERT ProcessDefinitionEntity[...]' with message 'Error updating database.  Cause: org.h2.jdbc.JdbcSQLException: Column "HISTORY_TTL_" not found; ...'

### DIFF
--- a/content/user-guide/process-engine/database.md
+++ b/content/user-guide/process-engine/database.md
@@ -158,8 +158,8 @@ The following properties can be set, regardless of whether you are using the JDB
 
 * `databaseType`: It's normally not necessary to specify this property as it is automatically analyzed from the database connection meta data. Should only be specified in case automatic detection fails. Possible values: {h2, mysql, oracle, postgres, mssql, db2, mariadb}. This setting will determine which create/drop scripts and queries will be used. See the 'supported databases' section for an overview of which types are supported.</li>
 * `databaseSchemaUpdate`: Allows to set the strategy to handle the database schema on process engine boot and shutdown.
-  * `false` (default): Checks the version of the DB schema against the library when the process engine is being created and throws an exception if the versions don't match.
-  * `true`: Upon building the process engine, a check is performed and an update of the schema is performed if necessary. If the schema doesn't exist, it is created.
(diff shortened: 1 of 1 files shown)


sdorokhova:
Hi @falko ,

are you sure that the default value for 'databaseSchemaUpdate' is true? From what I see in code the default is false, except for the StandaloneInMemProcessEngineConfiguration which defines default as "create_drop". 

Cheers,
Svetlana

falko:
Hi @sdorokhova ,

it's a very interesting question which "default" we consider in the documentation. Only a few people instantiate the configuration classes directly. In our distributions, it is always set to true. True would make sense as a default if the behavior is as I described it this PR. Apart from the default value discussion, can you confirm that my new description is in line with the current code?

Cheers,
Falko

sdorokhova:
Hi @falko ,
you're totally right about the `default` values. I agree that values provided in distributions must be considered as default ones.

As for the content of your changes, I like the changes in `true` value - old description was incorrect.

I'm not 100% sure about changes in `false` value description. I agree that the version of the database is not checked, but I see in code, that still some checks happening, which can throw an exception:
https://github.com/camunda/camunda-bpm-platform/blob/319d7eb2bc2d8f7f8fc10970aa38a28e678346a0/engine/src/main/java/org/camunda/bpm/engine/impl/db/sql/DbSqlSession.java#L295-L326

E.g. when first the engine was started and the DB was created with `cmmnEnabled=false`, the database would not have CMMN tables, and then, let's imagine, the configuration parameter was changed on `true` and engine restarted. In this case the exception might be thrown on startup.
But I would say that we can ignore this fact and left the simplified description in the docs, as probably nobody uses Camunda this way.

I will merge your changes. If we decide to be more precise in `false` case, we can improve it in next iteration.

Cheers,
Svetlana

PR #140, opened 2017-08-22
add right-click to delete something

ingorichtsmeier:
Add a hint to the right click in the cell to get rid of a rule or an input column.


ThorbenLindhauer:
Thanks Ingo!

PR #142, opened 2017-08-23
Corrected Process Search: String variable

GitMensch:
If I search for `customerId = "My-ID"` there is no process found. If I remove the quotes (and therefore do what filter-examples/var-string.png shows) the process is found.


yanavasileva:
Hello,

Best regards,
Yana

PR #147, opened 2017-10-20
Add example for Process Instance Restart REST API

benhoffmann:
The existing example is not very intuitive because normally one only uses one of the two options (processInstanceId or historicProcessInstanceQuery). I added another example to better show the different ways of finding the right instance.

### DIFF
--- a/content/reference/rest/process-definition/post-restart-process-instance-sync.md
+++ b/content/reference/rest/process-definition/post-restart-process-instance-sync.md
@@ -157,13 +157,13 @@ Request Body:
 }
 ```
 
-## Restarting one or more Process Instances using a historicProcessInstanceQuery:
-
 ### Response
 
 Status 204. No content.
 
-## Request
+## Restarting one or more Process Instances using a historicProcessInstanceQuery:
+
+### Request
 
 POST `/process-definition/aProcessDefinitionId/restart`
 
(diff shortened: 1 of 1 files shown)


content/reference/rest/process-definition/post-restart-process-instance-sync.md
```diff
@@ -151,6 +153,32 @@ Request Body:
   ],
   "initialVariables" : true,
   "skipCustomListeners" : true,
+  "withoutBusinessKey" : true
+}
+```
+
+## Restarting one or more Process Instances using a historicProcessInstanceQuery:
+
+### Response
```
pazolka: This section should be related to the first example. Please adjust it.

pazolka:
Please consider applying similar changes to the [async case. ](https://docs.camunda.org/manual/7.7/reference/rest/process-definition/post-restart-process-instance-async/) Also, follow my [hint. ](https://github.com/camunda/camunda-docs-manual/pull/147/commits/9b52d56ff47aef31f83face91466d453533f01d2#diff-5a5fed51bb28a0b6236a68836c36a40aR162)

benhoffmann:
Something got mixed up with the headers in my first commit, sorry for that! I fixed it.

PR #155, opened 2017-12-22
Update post-signal.md

benhoffmann:
According to the BPMN spec, signals are purely used in a broadcast manner. There might be a reason why we can signal single executions with our APIs, but the example and documentation should reflect the normal usage instead of some rarely used special case.


tasso94:
Thank you!

Cheers,
Tassilo

PR #156, opened 2018-01-05
Update get-form-variables.md

felix-mueller:
This endpoint is used in the camunda tasklist for embedded forms, too. I believe we should not mention that this applies only for generated forms - or do I misunderstand the brackets?
It is possible to retrieve variables by name by this endpoint even if they are not defined in the BPMN.

### DIFF
--- a/content/reference/rest/task/get-form-variables.md
+++ b/content/reference/rest/task/get-form-variables.md
@@ -12,8 +12,7 @@ menu:
 
 ---
 
-Retrieves the form variables for a task (only if they are defined via the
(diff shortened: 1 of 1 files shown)


yanavasileva:
Hi Felix,

My suggestion is to add the embedded task forms in the brackets too and not to delete them.
That way we will be more precise. What do you think?

Best regards,
Yana

yanavasileva:
Hi Felix,

I merged the change.

Thanks and best regards,
Yana

PR #157, opened 2018-01-10
Did not describe the behaviour properly

Tristan-ow:
this is not the time that the lock duration is extended but the total expiration time since the first lock

### DIFF
--- a/content/reference/rest/external-task/post-extend-lock.md
+++ b/content/reference/rest/external-task/post-extend-lock.md
@@ -45,7 +45,7 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>newDuration</td>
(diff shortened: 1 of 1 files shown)


sdorokhova:
Hi @Tristan-ow ,
thank you for noticing. I agree, that the docs don't explain the behaviour well enough, but I'm afraid that your understanding of the parameter is not correct. It actually means new lock duration from the current moment of time (the moment when `extendLock` method was called). 

I would formulate it like this:
"An amount of time (in milliseconds). This is the new lock duration starting from the current moment."

Can you give your feedback? Is it more understandable like this?

sdorokhova:
Pushed adjusted text in commit 289f64d

PR #169, opened 2018-03-31
Note on dangers of field injection into Spring beans

Infeligo:
As discussed in the forum [1], docs should clarify that the usage of field injection with Spring Beans via `delegateExpression` option is not a good idea, unless you know exactly what you are doing.

1. https://forum.camunda.org/t/field-injection-into-spring-beans-bad-idea/6266


koevskinikola:
Hey @Infeligo ,

I reviewed your pull request. I had to change the commit message to conform to the project's guidelines and fixed a typo as well. I also ported it to the master branch.

Thank you for the contribution!

Cheers,
Nikola

PR #173, opened 2018-04-27
mention the jdbcBatchProcessing setting

ingorichtsmeier:
Some customers ask for this, because they get database exceptions after upgrading to 7.8.

And the documentation is hidden in a bunch of other settings. It is not clear to them, where to set the property.


yanavasileva:
Hi Ingo,

As promised I merged the changes.

Best regards,
Yana

PR #175, opened 2018-04-28
doc(rest): clarify query parameter in history activity instance query

hongyuan1306:
The query parameter of REST history activity instance API could use a little more explanation. Otherwise I would expect a query with

> {"canceled": false}

to only return activity instances with property `canceled` value `false`

See forum thread
https://forum.camunda.org/t/rest-can-not-query-for-not-canceled-activity-instances/7044


koevskinikola:
Hey @hongyuan1306 ,

Once again, thanks for the contribution!

I merged the commit related to this pull request (dd6f147) with the master branch.

However, I noticed that there are 3 other commits in this pull request that are not related to the problem you were contributing to. This is not a good practice, since this way, we don't have a clear match of a defined problem with your complete solution/contribution.

Best,
Nikola

PR #180, opened 2018-05-31
fix spelling error in user-assignment.md

WilliamEddy:


yanavasileva:
Hi Kurtis,

Thank you for the contribution.
I merged the change as you can see in the above commit (f494c74).

Best regards,
Yana

PR #182, opened 2018-06-14
Fixed URL in create case instance response

pieter-vincken-agilians:
Changed the URL in the example response for creating a case instance from `process-instance` to the proper `case-instance` resource.


yanavasileva:
Hi Pieter,

Thanks for the contribution.
I merged the change (85a653d).

Best regards,
Yana

PR #184, opened 2018-06-26
docs(rest): Add the property suspended to task query

hongyuan1306:
Add the `suspended` property to docs for REST task query following the discussion https://forum.camunda.org/t/the-suspended-property-of-task/7781.


yanavasileva:
Hi @hongyuan1306,

I fixed the example's response also and merged the changes.

Best regards,
Yana

PR #188, opened 2018-07-29
Migrate docs to compile with current hugo

nikku:
Right now our docs do only compile with a custom made version, two year old version of HUGO that needs to be manually installed and wired.

The reason for this change was the introduction of a `cachedPartial` function which is built into `hugo >= 0.30` as `partialCached`, cf. https://github.com/camunda/camunda-docs-manual/commit/d7970ce083de16bd924971284964bdf03072bc49.

This PR updates the docs to be built with modern versions of hugo (`>= 0.30.0`). The changes in a nutshell are:

* use sections
* use `partialCached`
* fix baseurl
* fix absolute links to use `ref` instead of `relref`
* silence hugo warning by adding required boilerplate

Trace of building with `hugo@0.45.1`:

```bash
> hugo 
                                                                                                                                                                                     
                   | EN                                                                                                                                                              
+------------------+-----+
  Pages            | 994  
  Paginator pages  |   0  
  Non-page files   | 518  
  Static files     |  90  
  Processed images |   0  
  Aliases          |   0  
  Sitemaps         |   1                                                                                                                                                             
  Cleaned          |   0                                                                                                                                                             

Total in 807 ms                          
```


meyerdan:
Thank you @nikku. We should do this sooner than later. There is at least two things that come to mind

* we should do it on every branch
* the CI infrastructure needs to be mirgrated to that the new hugo version is provided to the builds. See [issue](https://app.camunda.com/jira/browse/SRE-64)

I will talk to @AndreaGiardini; SRE team currently has high workload, we will see when we can factor it in.

nikku:
It is a huge change and someone should help me verify that I did not accidentally break stuff before we port it to all branches.

nikku:
Output of `hugo@1.16-camunda-custom` building the site:

```bash
> hugo
Started building site
0 draft content
0 future content
841 pages created
518 non-page files copied
0 paginator pages created
0 tags created
0 categories created
in 15011 ms
```

It looks like the new version of hugo picks up more pages to be built.

We should verify if that is actually a case or if recent versions of hugo just count differently.

AndreaGiardini:
I had a quick look to this PR. Here there are a couple of comments:

 - **Number of pages generated**

As @nikku pointed out, from the summary printed by hugo it seems like the new version is generating way more pages that the previous version. I think it is just a matter of different counting:
```
➜ find new-hugo -type f | wc -l # Hugo v0.46
1603

➜ find old-hugo -type f | wc -l # Hugo custom
1604
```
The new hugo **does not generate** a 404 page. This should be fixed.

- **Page diff**

Almost all the pages have differences compared to the original version. What I noticed is that sometime the change is something like this:

**old hugo**
`<script type="text/javascript" src="/manual/develop/js/docs.js"></script>`
**new hugo**
`<script type="text/javascript" src="https://docs.camunda.org/manual/develop/js/docs.js"></script>`

Honestly I am not familiar enought with web development to know what is better :) I will let somebody else more knowledgeable than me make the choice. It looks like this is [configurable](https://gohugo.io/content-management/urls/#canonicalization)

One doubt that i have in my mind is... if we move Urls to the new format, every local deployment will reference the JS files from the live version or the local one?

---

I think there is more to investigate here... I see that some of the urls changed and this might require some redirection-magic or finding the correct hugo settings (if they exist) to keep the structure of the website intact. I will investigate more soon, in the meanwhile it would be nice to fix the first point and have your opinion about the second one :)

nikku:
Could you share the output of the two commands:

```
➜ find new-hugo -type f | wc -l # Hugo v0.46
1603
```

```
➜ find old-hugo -type f | wc -l # Hugo custom
1604
```

?

AndreaGiardini:
I've probably done something wrong in my previous `find` command since I am getting different results now...

I have noticed that the new hugo creates an rss feed for every subsection. In total it generates 153 rss feeds, instead the old one was generating only 10

Is this something that we want / care about ?

nikku:
Latest output when building from this branch:

```plain
                   | EN                                                                  
+------------------+-----+
  Pages            | 845
  Paginator pages  |   0
  Non-page files   | 518
  Static files     |  90
  Processed images |   0
  Aliases          |   0
  Sitemaps         |   1
  Cleaned          |   0
```

cmur2:
I will close this PR since the migration is done now for all branches.

PR #190, opened 2018-08-07
clarify on client details

ingorichtsmeier:
I added some findings of the external client behavior. They should answer some questions I got from customers and prospects recently.


yanavasileva:
Hi Ingo,

I merged the change (c185fa5).

Best regards,
Yana

PR #194, opened 2018-09-20
Document limitations of BpmnError

Infeligo:
I wanted to clarify the limitations of throwing BpmnError's from execution listeners. I always treated execution listeners just like any other delegation code and was surprised that BpmnErrors thrown from them were not caught by the boundary event, but turned into incidents. I found out that there is a corresponding [feature request](https://app.camunda.com/jira/browse/CAM-5399). Until it is implemented, it is good to have a warning about it in the docs.

### DIFF
--- a/content/user-guide/process-engine/delegation-code.md
+++ b/content/user-guide/process-engine/delegation-code.md
@@ -480,5 +480,11 @@ Throwing a `BpmnError` in the delegation code behaves like modeling an error end
 
 {{< /note >}}
 
+{{< note title="" class="warning" >}}
+
+Beware that currently throwing `BpmnError`'s from execution listeners (but not Java delegates) has limitations. The error will be treated as a runtime exception and will not be matched using error code by the boundary event.
+
+{{< /note >}}
+
 [script-sources]: {{< relref "user-guide/process-engine/scripting.md#script-source" >}}
 [camunda-script]: {{< relref "reference/bpmn20/custom-extensions/extension-elements.md#camunda-script" >}}


tasso94:
Hi Mihhail,

thank you for your contribution. Unfortunately, we've decided not to merge your contribution. 

And here is why: Even if there exists a feature request, it has not been said that the current behavior is unexpected. Both approaches are valid and we haven't decided yet which way to go on this topic. Merging your pull request to our documentation would mean, that it is planned to add this feature at one point in the future and therefore the current behavior is not intended.

Please feel free to reopen this issue if you have any further questions or thoughts.

Cheers,
Tassilo

Infeligo:
I would argue that the current behavior is unexpected according to the [Delegation Code](https://docs.camunda.org/manual/7.9/user-guide/process-engine/delegation-code/#throw-bpmn-errors-from-delegation-code) section of the documentation. Here is my breakdown:

In the beginning, the page states:

> There are different types of Delegation Code:
> - Java Delegates...
> - Execution Listeners...
> - Task Listeners...

It basically states that Execution Listeners are a type of Delegation Code. Then, in the end there is a subsection titled "Throw BPMN Errors from Delegation Code" which in a note states that:

> Throwing a BpmnError in the delegation code behaves like modeling an error end event.

It never states that the above applies only to Java Delegates. Therefore, it is unexpected when this approach suddenly doesn't work with Execution Listener.

My referring to the that feature request just intended (poorly) to say that other people also expect different behavior. In the end, you can decide not to implement that feature and that would be totally fine. The question is only about the clarity of the documentation.

PR #197, opened 2018-10-05
fix(links): change malformed or mistyped links to intended target

cmur2:
These were found using an automated tool and need to be changed for followup tasks.


koevskinikola:
Hi @cmur2,

Just merged your PR. Thanks!

Cheers,
Nikola

PR #207, opened 2019-01-10
chore(user-guide): fix typo

tobiasschaefer:
I chose 7.10 as a base for the merge request. Would "master" be better? I'd be happy to get a short explanation.

Please backport if needed.


tmetzke:
Hi @tobiasschaefer ,

thank you for the contribution.

The "master" would usually be a better base for the pull request in general as it might have already been fixed there in the meantime for example. In this case here it doesn't make that big of a difference since you are the first to fix it. Thanks again for that.

I merged it into master.

PR #214, opened 2019-01-25
Update call-activity.md

volkergersabeck:
in one case it should be calledElementVersionTag instead of just calledElementVersion


tmetzke:
Thanks Volker, will merge it to master after our alpha release this week.

Regards
Tobias

PR #220, opened 2019-02-19
fix(data-formats): fix links to JSON/XML variables

falko:


tmetzke:
Thank you for spotting this, Falko!
Merged it into master.

Best,
Tobias

PR #227, opened 2019-04-12
FIX typo: signalRef is being use for the message definition

divineforest:


tmetzke:
Hi @divineforest,

thanks for spotting this one!
Merged it into master.

Best,
Tobias

PR #233, opened 2019-05-29
[CAM-10381] read embedded form from the root context

ingorichtsmeier:
A reference from Thorbens interesting forum post: https://forum.camunda.org/t/embedded-forms/9941/2?u=ingo_richtsmeier.

Worth to be mentioned here.

### DIFF
--- a/content/user-guide/task-forms/_index.md
+++ b/content/user-guide/task-forms/_index.md
@@ -57,10 +57,11 @@ To add an embedded form to your application, simply create an HTML file and refe
 </form>
 ```
 
-The file containing the form can be referenced in two ways:
+The file containing the form can be referenced in three ways:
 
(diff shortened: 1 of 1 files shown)


tasso94:
Hi Ingo,

thanks for reaching out to us with this pull request.

The next step is to investigate a bit further if our test coverage is already good enough to guarantee that this behavior works as expected.

Stay tuned!

Cheers,
Tassilo

mboskamp:
Hi Ingo,

I have reviewed your changes. We have decided to not merge it yet, as we should test this properly before oficially documenting the third technique of loading task forms.
We will work on this first and then re-evaluate this PR.

You can track the progress of this with the linked CAM issue.

Cheers,
Miklas

PR #236, opened 2019-06-04
changed default path to /rest from /rest/engine

patrick-werner:
currently the default path mentioned here is /rest/engine, changed it to the correct value /rest


tmetzke:
Hi @patrick-werner ,

thanks a lot for spotting this!
I merged it into branches 7.11 and master.

Best,
Tobias

PR #289, opened 2019-09-09
[CAM-9855]: Update REST docs for update job due date endpoint

mboskamp:
Related to CAM-9855

### DIFF
--- a/content/reference/bpmn20/events/timer-events.md
+++ b/content/reference/bpmn20/events/timer-events.md
@@ -84,7 +84,7 @@ A timer's repeat cycle can be controlled via the [REST API](https://docs.camunda
 managementService.setJobDuedate(String jobId, Date newDuedate)
 ```
 
-Changes to one timer instance do not automatically affect subsequent timer instances. Given a recurring timer, that produces a timer event every 30 minutes. If the due date of one timer is changed (e.g. +15minutes) it would be executed 45 minutes after the previous timer. However, the following timer would follow the old pattern and would execute 15 minutes after the changed timer.
+Changes to one timer instance do not automatically affect subsequent timer instances. For example, a certain recurring timer produces a timer event every 30 minutes. If the due date of one timer event is changed (e.g. +15minutes), it would be executed 45 minutes after the previous timer. However, the following timer would follow the old pattern and would execute 15 minutes after the changed timer.
 
 If the desired outcome is to calculate the due dates of all subsequent timers based on the changes made, then it is possible to pass a `cascade` flag (when using the [REST API](https://docs.camunda.org/manual/latest/reference/rest/job/put-set-job-duedate/)) or use the following Java API method:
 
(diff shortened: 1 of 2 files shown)


koevskinikola:
I think we should also add a paragraph explaining that offsetting the Timer DueDate is possible, and what the behaviour is for multiple timers here: https://docs.camunda.org/manual/7.11/reference/bpmn20/events/timer-events/#time-cycle.

content/reference/bpmn20/events/timer-events.md
```diff
@@ -77,6 +77,22 @@ Note: The first symbol denotes seconds, not minutes as in normal Unix cron.
 
 The recurring time duration option is better suited for handling relative timers, which are calculated in respect to some particular point in time (e.g., the time when a user task was started), while cron expressions can handle absolute timers - which is particularly useful for timer start events.
 
+### Modify a Time Cycle
+A timer's repeat cycle can be controlled via the [REST API](https://docs.camunda.org/manual/latest/reference/rest/job/put-set-job-duedate/) or by calling the ManagementService. By setting the due date of a timer, is possible to change the point in time when a timer is executed.
```
koevskinikola:
```suggestion
A timer's repeat cycle can be controlled via the [REST API](https://docs.camunda.org/manual/latest/reference/rest/job/put-set-job-duedate/) or by calling the ManagementService. By setting the due date of a timer, it is possible to change the point in time when a timer is executed. 
```

content/reference/bpmn20/events/timer-events.md
```diff
@@ -77,6 +77,22 @@ Note: The first symbol denotes seconds, not minutes as in normal Unix cron.
...
 The recurring time duration option is better suited for handling relative timers, which are calculated in respect to some particular point in time (e.g., the time when a user task was started), while cron expressions can handle absolute timers - which is particularly useful for timer start events.
 
+### Modify a Time Cycle
+A timer's repeat cycle can be controlled via the [REST API](https://docs.camunda.org/manual/latest/reference/rest/job/put-set-job-duedate/) or by calling the ManagementService. By setting the due date of a timer, is possible to change the point in time when a timer is executed. 
+
+```java
+managementService.setJobDuedate(String jobId, Date newDuedate)
+```
+
+Changes to one timer instance do not automatically affect subsequent timer instances. Given a recurring timer, that produces a timer event every 30 minutes. If the due date of one timer is changed (e.g. +15minutes) it would be executed 45 minutes after the previous timer. However, the following timer would follow the old pattern and would execute 15 minutes after the changed timer.
```
koevskinikola:
```suggestion
Changes to one timer instance do not automatically affect subsequent timer instances. For example, a certain recurring timer produces a timer event every 30 minutes. If the due date of one timer event is changed (e.g. +15minutes), it would be executed 45 minutes after the previous timer. However, the following timer would follow the old pattern and would execute 15 minutes after the changed timer.
```

PR #290, opened 2019-09-12
[CAM-10444] docs(variables in rest api): sample request for Object type variable

garima-camunda:
related to CAM-10444


yanavasileva:
Merged with (2828cb8).

PR #305, opened 2019-10-08
CAM-10876: Update process instance restart REST API docs

superbeagle:
Updated process instance restart REST API docs to indicate they can be used to restart normally completed process instances. Also mentioned it creates a new instance based on the original instance information.


yanavasileva:
Hi Joe,

I will merge it tomorrow. Do you think that it should be backported as well? (for example 7.11)

Best regards,
Yana

superbeagle:
Hi Yana,
     Thanks, yeah it might be appropriate to backport it as well.

Joe

PR #318, opened 2019-10-16
CAM-10899: remove retired environments

ThorbenLindhauer:
https://app.camunda.com/jira/browse/CAM-10899

### DIFF
--- a/content/installation/full/jboss/_index.md
+++ b/content/installation/full/jboss/_index.md
@@ -12,6 +12,6 @@ menu:
 
 ---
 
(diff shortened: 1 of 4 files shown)


ThorbenLindhauer:
Thanks for the review. I pushed a commit with changes. I decided against merging the JBoss EAP 6 and Wildfly installation guides, because they require slightly different setup of the subsystem and I believe it is easier to follow if they are clearly separated. For the hint `SpingProcessApplication`, I don't know if this applies to recent Wildfly versions as well and couldn't find the root cause for this limitation in a quick search. I decided to change the hint to JBoss EAP 6.x for now.

PR #367, opened 2019-11-13
docs(rest): add parse exception description in error hadling overview

yanavasileva:
[CAM-10771](https://app.camunda.com/jira/browse/CAM-10771)


ThorbenLindhauer:
Merged via https://github.com/camunda/camunda-docs-manual/commit/4d1376e37b93a42983180003ec32af10df861e4b. I made some minor corrections and included the example also directly on the `POST /deployment/create` since its the only place where we returns this exception.

PR #379, opened 2019-11-19
[CAM-11112] docs(security): adjust type validation paragraph

tmetzke:
[CAM-11112](https://app.camunda.com/jira/browse/CAM-11112)

related to CAM-11112


tmetzke:
Great motivation, very practical angle, I like it! 
Thanks for giving this a thorough thought.
I would like to clarify that the custom classes/packages only extend the default lists and not override them. Other than that, impressive description. :)

PR #413, opened 2019-12-18
CAM-11158 Document property to disable history cleanup per node

tasso94:
[CAM-11158](https://app.camunda.com/jira/browse/CAM-11158)


yanavasileva:
Merged with (0bbe471).

PR #446, opened 2020-02-13
CAM-11394: Updates on skipped versions

superbeagle:
[CAM-11394](https://app.camunda.com/jira/browse/CAM-11394)

Added clarifications on how to 'skip' minor versions by indicating database alteration scripts from every minor version need to be executed. In a rolling update, every node has to be updated before the update to the next minor version can be executed.

### DIFF
--- a/content/update/minor/_index.md
+++ b/content/update/minor/_index.md
@@ -15,4 +15,8 @@ menu:
 These documents guide you through the process of updating your application or server installation from one Camunda minor version to the other.
 This applies to any updates of the version number "after the first dot", example: `7.3` to `7.4`.
 
+{{< note title="Skipping Minor Versions" class="warning" >}}
+It is possible to skip minor versions when updating your installation since application updates are cumulative. However, the database alteration scripts are NOT cumulative and need to be executed in version order and must include any skipped versions.
+{{< /note >}}
+
 There is a dedicated update guide for each version:
--- a/content/update/minor/_index.md
+++ b/content/update/minor/_index.md
@@ -15,8 +15,8 @@ menu:
 These documents guide you through the process of updating your application or server installation from one Camunda minor version to the other.
 This applies to any updates of the version number "after the first dot", example: `7.3` to `7.4`.
 
-{{< note title="Skipping Minor Versions" class="warning" >}}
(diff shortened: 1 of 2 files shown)


yanavasileva:
Hi Joe,

We will have a look to merge the proposal.
(Sorry for the referenced commit, I messed up the commit message there.)

Best regards,
Yana

ThorbenLindhauer:
Hey Joe,

Thanks for contribution. I think it is generally a good idea to make people aware that they do not have to update the Camunda libraries to every in-between version. However, we also see the risk that people may misunderstand the "Skipping Minor Versions" hint such that they will skip the migration guide for some versions entirely (except the DB part). We should try to make it more clear that the skipping only applies to the library update. If that is not so easy, I would rather not have this hint, as the current state causes confusion but no harm, whereas people skipping important parts of the migration guide can cause severe issues in their systems.

What do you think?

Cheers,
Thorben

superbeagle:
Hi Thorben,
     I certainly understand the challenges with potential confusion. I kept seeing this question being asked in different channels which was my original motivation for the update. I put the update in a warning box in the hopes of highlighting it. Maybe make it part of each update?

Joe

superbeagle:
or maybe a rewording of 'skip'

superbeagle:
Something like this?
{{< note title="Do I need to apply every minor version if I missed a few?" class="warning" >}}
If you need to apply multiple minor versions you MUST execute the database alteration scripts in minor version order as they are NOT cumulative. Application library updates are cumulative and don't require sequential minor version updates. You can simply apply your target minor version application libraries.
{{< /note >}}

ThorbenLindhauer:
I like the version in your last comment. Maybe we can add an extra sentence that the migration guide for every minor version should be checked.

superbeagle:
Will do....so once we come up with something should I publish again on this branch and create another pull request? Or use this one?

ThorbenLindhauer:
You can just add another commit to this branch. I will squash them on the merge.

superbeagle:
Will do and will add another commit later today

ThorbenLindhauer:
I merged your contribution via https://github.com/camunda/camunda-docs-manual/commit/590e46866f349e0fc895383d782fe712814c1edd. I changed the explanation why the migration guide should be consulted slightly, and I made the commit message comply to our commit message style. Thanks for your work!

PR #453, opened 2020-02-26
Set scope=test on Camunda BPM Assert Maven dependencies

falko:
[](https://app.camunda.com/jira/browse/)

Camunda BPM Assert and AssertJ are test tools and should therefore not be shipped to production. Setting the scope to test prevents users that copy the dependencies from accidentally leaking those test jars into their production wars.


mboskamp:
Thanks @falko, I will merge the change.

PR #479, opened 2020-03-23
CAM-11531: correct message documentation to reflect the implementation

mboskamp:
[CAM-11531](https://app.camunda.com/jira/browse/CAM-11531)

### DIFF
--- a/content/reference/rest/message/post-message.md
+++ b/content/reference/rest/message/post-message.md
@@ -190,7 +190,8 @@ Request Body:
       },
       "processVariables" : {
         "aVariable" : {"value" : "aNewValue", "type": "String", 
-                        "valueInfo" : { "transient" : true } },
+                        "valueInfo" : { "transient" : true }
+                      },
         "anotherVariable" : {"value" : true, "type": "Boolean"}
       }
     }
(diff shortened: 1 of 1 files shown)


content/reference/rest/message/post-message.md
```diff
@@ -43,15 +43,15 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>tenantId</td>
-    <td>Used to correlate the message for a tenant with the given id. Will only correlate to executions and process definitions which belong to the tenant.</td>
+    <td>Used to correlate the message for a tenant with the given id. Will only correlate to executions and process definitions which belong to the tenant. Must not be supplied in conjunction with a <code>withoutTenantId</code>.</td>
   </tr>
   <tr>
     <td>withoutTenantId</td>
-    <td>A Boolean value that indicates whether the message should only be correlated to executions and process definitions which belong to no tenant or not. Value may only be <code>true</code>, as <code>false</code> is the default behavior.</td>
+    <td>A Boolean value that indicates whether the message should only be correlated to executions and process definitions which belong to no tenant or not. Value may only be <code>true</code>, as <code>false</code> is the default behavior. Must not be supplied in conjunction with a <code>withoutTenantId</code>.</td>
```
koevskinikola: I think it should be:
```suggestion
    <td>A Boolean value that indicates whether the message should only be correlated to executions and process definitions which belong to no tenant or not. Value may only be <code>true</code>, as <code>false</code> is the default behavior. Must not be supplied in conjunction with a <code>tenantId</code>.</td>
```

mboskamp: You are right. Shame on me :(

PR #508, opened 2020-04-02
Bugfix cam 11717

aravindhrs:
Variables.objectValue(order) will return type of org.camunda.bpm.engine.variable.value.builder.ObjectValueBuilder which is not a type of org.camunda.bpm.engine.variable.value.TypedValue.

The code should be like, Variables.objectValue(order).create() will return the TypedValue.

com.example.Order order = new com.example.Order();
    
VariableMap variables = 
Variables.createVariables()
            .putValueTyped("order", Variables.objectValue(order).create())
            .putValueTyped("string", Variables.stringValue("a string value"))
            .putValueTyped("stringTransient", Variables.stringValue("foobar", true));
runtimeService.setVariablesLocal(execution.getId(), "order", variables);


marstamm:
Hey @aravindhrs ,

first of, thank you for your contribution.
Please understand that we might not respond to pull requests right away. We will usually respond within a week.

Cheers
Martin

PR #516, opened 2020-04-10
CAM-11492 : get list of topic names

epollum:
feat(api): get list of topic names


epollum:
@tasso94 , here is the PR for the documentation related to https://github.com/camunda/camunda-bpm-platform/pull/779

PR #528, opened 2020-04-28
when no input values are supplied when evaluating

pradeepkodi15:
This is will help the user to understand behavior of rest API very clearly.

### DIFF
--- a/content/reference/rest/decision-definition/post-evaluate.md
+++ b/content/reference/rest/decision-definition/post-evaluate.md
@@ -16,7 +16,7 @@ menu:
 
 ---
 
-Evaluates a given decision and returns the result. The input values of the decision have to be supplied in the request body.
+Evaluates a given decision and returns the result. The input values of the decision have to be supplied in the request body. When no input values are supplied an exception will be raised.
 
 
 # Method
--- a/content/reference/rest/decision-definition/post-evaluate.md
+++ b/content/reference/rest/decision-definition/post-evaluate.md
@@ -16,7 +16,7 @@ menu:
 
 ---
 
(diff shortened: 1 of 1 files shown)


yanavasileva:
Hi @pradeepkodi15,

I just want to clarify why this will bring value to the documentation.
Have you checked the [response code ](https://github.com/camunda/camunda-docs-manual/blob/master/content/reference/rest/decision-definition/post-evaluate.md#response-codes) section?
Without providing the variables the evaluation will not make sense in general, maybe we can improve the response code section instead of the general description?

Please share your thoughts about it.

Best regards,
Yana

pradeepkodi15:
Hi @yanavasileva  I agree with your comment. I initially thought adding the text in the description would help new users to make sure to add all input variables in the request body.

As mentioned in other section Response code do not cover all possible error message due to error may arise due to different reasons.   

I am not sure how we can improve the response section?

Thanks and Regards,
Pradeep

yanavasileva:
Hi Pradeep,

I would suggest putting clarification to 500 response code: 
```
The decision could not be evaluated successfully, e.g. some of the input values are not provided but they are required. See the <a href="{{< ref "/reference/rest/overview/_index.md#error-handling" >}}">Introduction</a> for the error response format.
```
Link: https://github.com/camunda/camunda-docs-manual/pull/528/files#diff-d4c7b9c3ca84bb00386d5e34893045a5L105

If the user doesn't provide the input values, they will receive an error. Then they will check the response codes to check for more information and the description will give the hint to check the input values.

What do you think?

Best regards,
Yana

pradeepkodi15:
I changed as per your comment. Please check it. Thanks

yanavasileva:
Hi Pradeep,

Thank you for your contribution once again.

I changed the commit message and merged the change with (c18f549).

Best regards,
Yana

PR #534, opened 2020-04-29
Grammatical error

tiruis:
displays is changed to display

### DIFF
--- a/content/webapps/cockpit/dashboard.md
+++ b/content/webapps/cockpit/dashboard.md
@@ -30,7 +30,7 @@ Please note that this feature is only included in the enterprise edition of the 
 
 {{< img src="../img/dashboard-metrics.png" title="Cockpit Dashboard Metrics" >}}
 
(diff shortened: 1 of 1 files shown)


yanavasileva:
Hi @tiruis,

I think the third singular form of `display` verb is correctly used, as it's paired with `the metrics plugin` (third-person singular).

Hereby I am closing the ticket. Feel free to open it at any time in case of questions.

Best regards,
Yana

tiruis:
"the metrics plugin displays graphs"  in this statement the metrics plugin is third person singular  so verb "displays" is allowed.But again "graphs" plural is coming which is not allowed.
Actually the statement should be like "the metrics plugin displays graph"

Let me know your concern

yanavasileva:
Hi @tiruis,

> But again "graphs" plural is coming which is not allowed.
Unfortunately, I do not agree as such grammatical rule doesn't exist.
Once again thank you for your interest in improving our documentation, however, I don't think that any change should be made in this case.

Best regards,
Yana

PR #541, opened 2020-04-29
chore(run): add documentation about running Camunda BPM Run with Docker

mboskamp:
Related to CAM-11305

### DIFF
--- a/content/installation/docker.md
+++ b/content/installation/docker.md
@@ -25,6 +25,8 @@ docker pull camunda/camunda-bpm-platform:latest
 docker run -d --name camunda -p 8080:8080 camunda/camunda-bpm-platform:latest
 ```
 
+Please note that by default the Apache Tomcat distribution is used. For a guide on how to use one of the other distributions, see the [tag schema](https://github.com/camunda/docker-camunda-bpm-platform#supported-tagsreleases).
+
 ## Start Camunda BPM Run using Docker
 
 To start [Camunda BPM Run]({{< ref "/user-guide/camunda-bpm-run.md" >}}) execute the following commands:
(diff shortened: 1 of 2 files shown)


content/installation/docker.md
```diff
@@ -14,7 +14,25 @@ menu:
...
+
+## Start Camunda BPM using Docker
+
+To start the Camunda BPM platform execute the following commands:
+
+```shell
+docker pull camunda/camunda-bpm-platform:latest
+docker run -d --name camunda -p 8080:8080 camunda/camunda-bpm-platform:latest
+```
+
```
koevskinikola, line 27: Maybe we should specify that Apache Tomcat is used by default. Feel free to modify the suggestion:

```suggestion

Please note that, by default, the Apache Tomcat distribution is used. For how to use one of the other distributions, see the [tag schema](https://github.com/camunda/docker-camunda-bpm-platform#supported-tagsreleases).

```

content/user-guide/camunda-bpm-run.md
```diff
@@ -39,6 +39,18 @@ camunda-bpm-run
...
 
+## Starting Camunda BPM Run using Docker
+
+Run is also available as Docker image. It can be found on [GitHub](https://github.com/camunda/docker-camunda-bpm-platform) and [Docker Hub](https://hub.docker.com/r/camunda/camunda-bpm-platform/tags?page=1&name=run). To start the latest release of Camunda BPM Run execute the following commands:
+
+```shell
+docker pull camunda/camunda-bpm-platform:run-latest
+docker run -d --name camunda -p 8080:8080 camunda/camunda-bpm-platform:run-latest
+```
+After a few seconds Camunda will be available at `http://localhost:8080`.
```
koevskinikola, line 50: I think it would be better to link to the Run section of the Docker documentation here. That way we avoid duplicating content and make it easier to maintain.
```suggestion
Run is also available as a Docker image. Please see the Run section of the Camunda Docker documentation [here](/installation/docker.md#start-camunda-bpm-run-using-docker) for more details.
```

content/user-guide/camunda-bpm-run.md
```diff
@@ -39,6 +39,18 @@ camunda-bpm-run
...
+
+If you want to use the enterprise version, you can build it yourself with the steps described on [GitHub](https://github.com/camunda/docker-camunda-bpm-platform#build-a-enterprise-version).
```
koevskinikola, line 52: I would also move this to the EE section of the Docker docs since it's valid for all Docker images, and we currently don't state this information, so it would be very useful for EE users of all Docker images.

PR #545, opened 2020-05-01
Update business-rule-task.md: where hyper link in the given section is not working as expected

aloksingh25:
IN business rule task, link in the section "Implementing as an External Task" is borken. Trying to fix it here for the readers.


marstamm:
Hi @aloksingh25 ,

thank you for making us aware. However, you PR did not fix the problem. The Problem was caused by broken markdown syntax. I fixed it in the referenced commit.

Cheers
Martin

PR #546, opened 2020-05-02
Update hit-policy.md: Adding to consequence of no rule match for Unique Hit Policy

aloksingh25:
Adding to consequence of no rule match for Unique Hit Policy


aloksingh25:
Closing due to wrong doc version

PR #559, opened 2020-05-11
CAM-11527

epollum:
@yanavasileva ,

Here is the updated docs for the missing items found during CAM-11527


yanavasileva:
The added properties are valid for the count query as well, could you please add them there as well?

yanavasileva:
👍

yanavasileva:
Changed the commit message and merged with (51668e7).

PR #560, opened 2020-05-12
CAM-11929: Update hit-policy.md: Adding to emphasis if no rule matched for the given input in case of Unique hit policy

aloksingh25:
Adding to emphasis if no rule matched for the given input in case of Unique hit policy.

### DIFF
--- a/content/reference/dmn11/decision-table/hit-policy.md
+++ b/content/reference/dmn11/decision-table/hit-policy.md
@@ -71,12 +71,10 @@ multiple satisfied rules.
 
 ## Unique Hit Policy
 
-Only a single rule can be satisfied. The decision table result contains the
+Only a single rule can be satisfied or no rule at all. The decision table result contains the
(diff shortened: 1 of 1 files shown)


content/reference/dmn11/decision-table/hit-policy.md
```diff
@@ -76,6 +76,8 @@ output entries of the satisfied rule.
 
 If more than one rule is satisfied, the Unique hit policy is violated.
 
+Exactly one rule or no rules should match for a given input.
```
koevskinikola, line 79: I think this is redundant since the same is clearly stated in the line above:

>  Only a single rule can be satisfied. The decision table result contains the output entries of the satisfied rule.

@aloksingh25, what would be the benefit of adding this line? Wouldn't it be better to modify the line that I mention like: "Only a single rule can be satisfied, or no rule at all"?

Best.
Nikola

PR #564, opened 2020-05-14
chore(release): add entity relationship diagrams for 7.13

mboskamp:
Related to CAM-11846


mboskamp:
The `.mwb` files are used to generate the `.svg`s from the MySQL workbench tool. We should keep them for older versions in case we want to change the DB model and it's much handier to have them all on the latest branch. I would keep it that way and merge the ticket.

PR #570, opened 2020-05-18
Business rule task can be used as service task

pradeepkodi15:
Added Business rule task can also be used as a service task with different implementation approaches.


tasso94:
Thank you for your contribution!

PR #576, opened 2020-05-19
Keystroke and table

tiruis:
It will look good if we form a table with action against the key stroke for Keyboard Shortcuts.


tasso94:
Hi @tiruis,

Thanks for your contribution.

I made a tiny adjustment by renaming the table header `KeyStroke` to `Shortcut` and merged your changes with the master branch.

Best,
Tassilo

PR #592, opened 2020-06-11
spelling mistake documenation

pradeepkodi15:
spelling mistake documenation. it should be documentation.


mboskamp:
Thanks for the fix @pradeepkodi15, well spotted. :)

PR #593, opened 2020-06-11
Update event-based-gateway.md

vibhortayal:
Typo


mboskamp:
Thanks for the fix. Note that this will only fix the 7.5 branch of the docs. I see that it was fixed for 7.7+ so it is not necessary to port it to current branches.

PR #606, opened 2020-06-16
docs(webapps): add documentation about hsts header support

tasso94:
related to CAM-11602


tmetzke:
I added a small spelling correction as a new commit, please squash on/before merge

PR #609, opened 2020-06-17
CAM-11873 - feat(licenses): add html license book

marstamm:
related to CAM-11873


ThorbenLindhauer:
As discussed, closing the PR as the submitted document is for a snapshot version. Let's decide with legal if we want to a) generate the next markdown book with 7.14.0, b) do it with the next alpha c) do it for a previously released version (e.g. 7.13.0).

PR #634, opened 2020-07-14
docs: Mapping JSON ref - update link for Jackson docs

SaumyaJain15:
The current link for Jackson documentation is not available and needs to be replaced. Adding a github link instead of existing link.


mboskamp:
Hi @SaumyaJain15, thanks for fixing this. Nice catch!

PR #678, opened 2020-09-04
docs(rest): add batch id to variable instance endpoints

tasso94:
- Adds `batchIdIn` query criterion
- Adds `batchId` property to variable instance results

Related to CAM-12304


tmetzke (approved):
Looks good to me 👍

PR #689, opened 2020-09-14
docs(update): Add section about Cockpit Plugins

marstamm:
related to CAM-11443

### DIFF
--- a/content/update/minor/713-to-714/_index.md
+++ b/content/update/minor/713-to-714/_index.md
@@ -21,6 +21,7 @@ This document guides you through the update from Camunda BPM `7.13.x` to `7.14.0
 1. For developers: [Update to JQuery 3.5](#update-to-jquery-3-5)
 1. For developers: [Changes to Task Query and Historic Task Query behavior](#changes-to-task-query-and-historic-task-query-behavior)
 1. For developers: [New Engine Dependency - Connect](#new-engine-dependency-connect)
+1. For developers: [New Frontend Plugin System for Cockpit](#new-frontend-plugin-system-for-cockpit)
 
 This guide covers mandatory migration steps as well as optional considerations for the initial configuration of new functionality included in Camunda BPM 7.14.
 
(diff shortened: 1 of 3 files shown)


tasso94 (changes requested):
Looks good to me in general. However, we should also document the following points:
* The structure of `bpmnJs.additionalModules` changed
* We should document the implications that no preconfigured HTTP client is available:
   * How can I get the CSRF token?
   * How can I get the API paths?
   * Is there any recommendation what I should use (i.e., fetch)?
* The CSS classes of the header changed, so the ["Logo and Header Color"](https://docs.camunda.org/manual/develop/webapps/cockpit/extend/configuration/#logo-and-header-color) example does not work anymore
  * I know from many cases that users make use of customizing the header
  * We should document this change in the migration guide and update the example
* Can we bring back the infographic about "The Nature of a Cockpit Plugin"?
* ["Advanced Styles Customization"](https://docs.camunda.org/manual/develop/webapps/cockpit/extend/configuration/#advanced-styles-customization)
   * Can we update the whole paragraph? 
   * Can we mention in the migration guide that this is deprecated and will be removed soon?

marstamm:
Great catches, thank you @tasso94 !

### HTTP client:
 I see this as a guide to migrate using AngularJS, therefore I would not recommend `fetch` to the users. In the documentation and examples to write **new** Plugins, we already use `fetch`. If the User did not change the CSRF cookie name, the default `$http` provider will pick it up from the cookies.
I still added a paragraph about it in case they changed the cookie name or want to migrate away from AngularJS.

### "Advanced Styles Customization"
This might be difficult for the new build as it is a mix of scss and less - and the scss result has a hashed file name. I think we should remove the section completely and users should only use the `user-styles.css` to override the camunda styles

### Infographic
I never saw much value in the graphic... If you think it is important, I suggest we create a follow-up ticket for it

content/update/minor/713-to-714/_index.md
```diff
@@ -21,6 +21,8 @@ This document guides you through the update from Camunda BPM `7.13.x` to `7.14.0
 1. For developers: [Update to JQuery 3.5](#update-to-jquery-3-5)
 1. For developers: [Changes to Task Query and Historic Task Query behavior](#changes-to-task-query-and-historic-task-query-behavior)
 1. For developers: [New Engine Dependency - Connect](#new-engine-dependency-connect)
```
tasso94, line 23:
```suggestion
1. For developers: [New Engine Dependency - Connect](#new-engine-dependency-connect)
1. For developers: [Cockpit Style Customizations](#style-customizations)
```

content/update/minor/713-to-714/_index.md
```diff
@@ -165,6 +167,96 @@ Camunda Connect dependency has been added to the process engine (`camunda-engine
 In case you already have a [Connect]({{< ref "/reference/connect/_index.md#maven-coordinates" >}}) dependencies to some of your projects, please consider consolidating the version of them with one that comes as dependency with the engine. That will prevent inconsistencies on the system. Please note that the Connect process engine plugin is still an optional dependency.
 
 
+# Style Customizations
```
tasso94, line 170:
```suggestion
# Cockpit Style Customizations
```
Makes it consistent to the headlines below.

tasso94 (approved):
Looks good. Just fixed two small things.

PR #703, opened 2020-09-18
chore(telemetry): add license key data

tmetzke:
related to CAM-12379


yanavasileva (approved):
Already approving for merge as the adjustment is small.

PR #706, opened 2020-09-22
CAM-12097: chore(docs): create a single page for licence info

koevskinikola:
* Add a single page dedicated to the Camunda License usage and retrieval
  methods.
* Adjust the doc sections where License use is described to point to the
  dedicated License page.

Related to CAM-12097

### DIFF
--- a/content/user-guide/license-use.md
+++ b/content/user-guide/license-use.md
@@ -1,0 +1,132 @@
+---
+
+title: 'Camunda License Keys'
+weight: 220
+
+menu:
+  main:
+    identifier: "user-guide-license"
+    parent: "user-guide"
+
+---
+
+Some Camunda BPM features (e.g. enterprise plugins) require a license key. The license will be provided as a string by 
+the Camunda support team. The license mechanism has no impact on the engine or other runtime components. The following 
+section explains the various methods by which a Camunda license can be added to the Process Engine.
+
+
+# Deployment Scenarios and the License Key
+
+A Camunda license key should be applied per database.
+
+In a clustered scenario, where multiple engines on multiple nodes access a single database, the 
+license only needs to be activated once. When activated, a license is valid until the expiration 
+date or until you have deleted your database. The license key is valid for an unlimited amount of 
+engines.
+
+In a multi tenancy scenario, the license check will be performed for each engine with an own 
+database. Thus, you will be prompted to enter the license key separately for each engine.
+
+
+# License Key Pickup Methods
+
+## Through the Admin Webapp UI
+
+Any environment setup that uses the enterprise Camunda Admin webapp, can use the provided License key page to
+enter a valid license key. Please see the dedicated [Admin webapp]({{< ref "/webapps/admin/system-management.md#camunda-license-key" >}}) 
+docs section for more details.
+
+## Through the Java API
+
+This method can be used in any environment setup. The license key can be set via the 
+Java API by calling:
+
+```
+managementService.setLicenseKey(String licenseKey);
+```
+
+The managementService also offers methods to get and delete the license key from the database.
+The license key is stored in the `ACT_GE_BYTEARRAY` table. A reference to the license entry can 
+be found in the `ACT_GE_PROPERTY` table (`camunda-license-key-id`).
+
+## Through the Home Directory
+
+Another possibility is to put a file with the license key on the path `${user.home}/.camunda/license.txt`. 
+This is another method that can be applied in any environment setup. The license key will be 
+automatically loaded in the database table unless a valid license key is already present.
+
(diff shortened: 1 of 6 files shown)


content/user-guide/license-use.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Through the Admin Webapp UI
+
+Any environment setup that uses the enterprise Camunda Admin webapp, can use the provided License key page to
+enter a valid license key. Please see the dedicated [Admin webapp]({{< ref "/webapps/admin/system-management.md#camunda-license-key" >}}) 
+docs section for more details.
+
+## Through the Java API
+
+This method can be used in any scenario, i.e. environment setup. The license key can be set via the
```
mboskamp, line 41:
```suggestion
This method can be used in any scenario, e.g. environment setup. The license key can be set via the 
```
I think this should be e.g. instead of i.e., right?
e.g. = for example
i.e. = that is

content/user-guide/license-use.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+The managementService also offers methods to get and delete the license key from the database.
+The license key is stored in the `ACT_GE_BYTEARRAY` table. A reference to the license entry can 
+be found in the `ACT_GE_PROPERTY` table (`camunda-license-key-id`).
+
+## Through the Home Directory
+
+Another possibility is to put a file with the license key on the path `${user.home}/.camunda/license.txt`. 
+This is another method that can be applied in any scenario, i.e. environment setup. The license key will be
```
mboskamp, line 55: Same as above

content/user-guide/license-use.md
```diff
@@ -0,0 +1,132 @@
...
+
+The managementService also offers methods to get and delete the license key from the database.
+The license key is stored in the `ACT_GE_BYTEARRAY` table. A reference to the license entry can 
+be found in the `ACT_GE_PROPERTY` table (`camunda-license-key-id`).
+
+## Through the Home Directory
+
+Another possibility is to put a file with the license key on the path `${user.home}/.camunda/license.txt`. 
+This is another method that can be applied in any scenario, i.e. environment setup. The license key will be 
+automatically loaded in the database table unless a license key is already present.
```
mboskamp, line 56:
```suggestion
automatically loaded in the database table unless a valid license key is already present.
```

content/user-guide/license-use.md
```diff
@@ -0,0 +1,132 @@
...
+
+## With the Camunda Spring Boot Starter & Camunda Run
+
+Camunda Run, and other Spring Boot applications that use the Camunda Spring Boot Starter, can provide license
```
mboskamp, line 60:
```suggestion
Camunda Run, and other Spring Boot applications (including Camunda BPM RPA Bridge) that use the Camunda Spring Boot Starter, can be provided with license 
```

mboskamp (changes requested):

mboskamp:
Nice work @koevskinikola! I only have a few tiny improvement suggestions.

content/user-guide/license-use.md
koevskinikola: Will add it with the rest of the review hint implementations.

content/user-guide/license-use.md
koevskinikola: Same as above.

content/user-guide/license-use.md
koevskinikola: I'm using "environment setup" in the same way as "scenario" here. I think the usage of "i.e." is correct. 

Environment setup would be the combination of OS+AS+DB+JDK (and any other underlying technologies used to run the engine, e.g. containers, framework). 

Scenario is more specific, previously we have used to term do describe setting up the engine in a cluster topology, or with multi tenancy. 

I think they can be used interchangeably in the context of licenses, but if only one should be used, I would go with "environment setup" since it's more general.

What do you think @mboskamp ?

content/user-guide/license-use.md
koevskinikola: The same comment from above applies here.

content/user-guide/license-use.md
koevskinikola: I dug around our docs a bit, so I think my changes would fit nicely:

* We use `Deployment Scenarios` to describe the topology model. E.g. embedded engine, container-managed, standalone, cluster model or multi-tenancy model.
* We use environment (setup), to describe the combination of OS+AS+DB+JDK+framework. This fits with how a license can be added: directory in OS, Java API, Java classpath, URL or Webapp UI. They are all part of some environment.

mboskamp: Alright, makes sense.

mboskamp (approved):

PR #714, opened 2020-09-28
chore(env): add support for JBoss EAP 7.3

mboskamp:
Related to CAM-12019


koevskinikola (approved):
Looks good! 

We should also add the new version in the announcements page here: https://github.com/camunda/camunda-docs-static/blob/master/enterprise/content/announcement.md

PR #715, opened 2020-09-28
CAM-12520: explain what 'finished' in post-process-instance-query means

McAlm:
As the docs do not explain what finished as parameter actually means...

### DIFF
--- a/content/reference/rest/history/process-instance/post-process-instance-query.md
+++ b/content/reference/rest/history/process-instance/post-process-instance-query.md
@@ -113,7 +113,7 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>finished</td>
-    <td>Only include finished process instances. Value may only be <code>true</code>, as <code>false</code> is the default behavior.</td>
+    <td>Only include finished process instances. Value may only be <code>true</code>, as <code>false</code> is the default behavior. This flag includes all process instances that are completed or terminated</td>
   </tr>
   <tr>
     <td>unfinished</td>


koevskinikola:
Hi @McAlm,

Since we're in the middle of a release/code-freeze, I'll try to review the PR and assign it to a ticket soon, but this might be delayed until the release is done (October 13th). 

In the meantime, please sign the CLA since any contributor to the Camunda project needs to do this the first time they contribute.

Best,
Nikola

content/reference/rest/history/process-instance/post-process-instance-query.md
```diff
@@ -113,7 +113,7 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>finished</td>
-    <td>Only include finished process instances. Value may only be <code>true</code>, as <code>false</code> is the default behavior.</td>
+    <td>Only include finished process instances. Value may only be <code>true</code>, as <code>false</code> is the default behavior. This flag includes all process instances that are completed or terminated</td>
```
koevskinikola, line 116: Let's keep the possible/default values at the end. This is the pattern that we maintain in all descriptions.

```suggestion
    <td>Only include finished process instances. This flag includes all process instances that are completed or terminated. Value may only be <code>true</code>, as <code>false</code> is the default behavior.</td>
```

koevskinikola (changes requested):
Please also add the changes to the OpenAPI documentation here: 

https://github.com/camunda/camunda-bpm-platform/blob/master/engine-rest/engine-rest-openapi/src/main/templates/lib/commons/history-process-instance.ftl#L102

This way the content is the same everywhere.

koevskinikola:
Hey @McAlm,

Thanks for raising this issue. I modified your contribution according to my review hints and mirrored it to the rest of the Rest API endpoints that use the same property. I also added it to the OpenAPI docs.

Cheers,
Nikola

PR #752, opened 2020-10-15
chore(docs): refactor plugins page

marstamm:
* consistent spacing between content and next headings (Markdown only)
* remove duplicated label explanation in Process Definition Runtime Tab
* Display additional arguments in a list

related to CAM-12591

PR #753, opened 2020-10-15
chore(docs): refactor plugins page

marstamm:
* consistent spacing between content and next headings (Markdown only)
* remove duplicated label explanation in Process Definition Runtime Tab
* Display additional arguments in a list

related to CAM-12591


tmetzke (approved):
Come on, it is the same PR as this one, just on a different branch: https://github.com/camunda/camunda-docs-manual/pull/752

Are you trying to get a shirt 👕?

Looks all good 👍

PR #758, opened 2020-10-18
Optimize images

JeffResc:
This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 33%.


marstamm:
Hey @JeffResc ,

thank you for your contributions. I see that this would improve loading times, but don't think this is the correct place to implement the changes.

Many of the pictures are automatically created with every release, so the improvements would not persist. I therefore think that image compression should be automated during the build step with `hugo`.

I created CAM-12623 to track the issue. 
While I'm closing the PRs, we still appreciate your contribution. If you have any questions or further improvement ideas, feel free to approach us.

Cheers
Martin

PR #761, opened 2020-10-22
docs(plugin): add static resource serving info

marstamm:
related to CAM-12620

### DIFF
--- a/content/webapps/cockpit/extend/plugins.md
+++ b/content/webapps/cockpit/extend/plugins.md
@@ -67,7 +67,7 @@ As runtime relevant resource it defines
 
 1. a plugin main class
 2. a `META-INF/services` entry that publishes the plugin to Cockpit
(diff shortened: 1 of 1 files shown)


marstamm:
I'm not sure if the resources part of the file tree is still correct, as we describe the static Resources to reside in `/plugin-webapp/$plugin_id`. Is there another configuration where the current location `org/my/plugin/assets/app/app/` is used?

tasso94 (changes requested):
By default, the asset folder is `/plugin-webapp/$plugin_id`. It can be changed by overriding `AbstractAppPlugin#getAssetDirectory` [1]. Can we change the file tree structure to the default?

[1] https://github.com/camunda/camunda-bpm-webapp/blob/master/src/main/java/org/camunda/bpm/webapp/plugin/spi/impl/AbstractAppPlugin.java#L46-L48

tasso94 (approved):

PR #777, opened 2020-10-29
CAM-11528: fix(reference): fix batch Rest API docs

yanavasileva:
CAM-11528

### DIFF
--- a/content/reference/rest/batch/delete.md
+++ b/content/reference/rest/batch/delete.md
@@ -68,9 +68,9 @@ This method returns no content.
     <td>Request successful.</td>
   </tr>
   <tr>
-    <td>404</td>
+    <td>400</td>
     <td>application/json</td>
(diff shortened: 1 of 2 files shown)


content/reference/rest/batch/delete.md
```diff
@@ -68,9 +68,9 @@ This method returns no content.
     <td>Request successful.</td>
   </tr>
   <tr>
-    <td>404</td>
+    <td>400</td>
```
koevskinikola, line 71: @yanavasileva are you sure this fix is correct? Usually `404` is used for `Not Found`/`Does not exist` errors. We have it documented for several other endpoints as well (e.g. https://docs.camunda.org/manual/latest/reference/rest/external-task/post-complete/)

koevskinikola: Confirmed that this is not the case for Batch operations.

koevskinikola (approved):

PR #804, opened 2020-11-28
CAM-12679: fix Camunda BPM Run logging properties

silvanbrenner:
depending on: https://github.com/camunda/camunda-bpm-platform/pull/1156

Maybe we should also considering to remove the link to https://howtodoinjava.com/spring-boot2/logging/configure-logging-application-yml. Because there is also the old property described.

### DIFF
--- a/content/user-guide/camunda-bpm-run.md
+++ b/content/user-guide/camunda-bpm-run.md
@@ -284,7 +284,7 @@ For more information on logging configuration visit the [Spring Boot Logging Gui
       <td><code>-</code></td>
   </tr>
   <tr>
-      <td><code>.file</code></td>
+      <td><code>.file.name</code></td>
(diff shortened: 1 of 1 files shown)


mboskamp:
Hi @silvanbrenner,
I agree we should change the link showing the old property.
I would suggest using `https://docs.spring.io/spring-boot/docs/2.4.0/reference/html/spring-boot-features.html#boot-features-logging` instead. What do you think?

Cheers,
Miklas

silvanbrenner:
Hi @mboskamp,
Yes, thats a good idea to reference the offical documentation from Spring Boot.

mboskamp:
@silvanbrenner,
alright, I will take care of changing the link. Thanks again for making this contribution. 👍 

Cheers,
Miklas

PR #806, opened 2020-12-07
CAM-12833-Fixed wrong API description

mimaom:


content/reference/rest/job/post-query-count.md
```diff
@@ -38,7 +38,7 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>jobIds</td>
-    <td>Filter by a comma-separated list of job ids.</td>
+    <td>Filter by list of job ids. Must be a JSON array of Strings.</td>
```
yanavasileva, line 41:
```suggestion
    <td>Filter by a JSON array of job ids.</td>
```

content/reference/rest/job/post-query-count.md
```diff
@@ -50,7 +50,7 @@ A JSON object with the following properties:
   </tr>
   <tr>
     <td>processInstanceIds</td>
-    <td>Only select jobs which exist for the given comma-separated list of process instance ids.</td>
+    <td>Only select jobs which exist for the given list of process instance ids. Must be a JSON array of Strings.</td>
```
yanavasileva, line 53:
```suggestion
    <td>Only select jobs which exist for the given a JSON array of process instance ids.</td>
```

yanavasileva (commented):
Hi @mimaom,

Could you please have a look at my suggestion, I believe they will be sufficient as they follow the same pattern as other endpoints description. If you agree with adjustments, please apply them for both endpoints.

Best,
Yana

content/reference/rest/job/post-query-count.md
mimaom: Works for me - as long as it's clear. I used the "Must be a JSON array of String" because that is also used in other places of the REST API.

yanavasileva:
Hi @mimaom,

Once again thank you for your contribution.
I merged the PR with (1715aca).

Best,
Yana

PR #816, opened 2020-12-15
fix(rest): minor improvements to variable-instance

yT0n1:
Related to CAM-11549


yanavasileva (approved):
Thank you for considering the hints. Please remove the sort mention in the responses of the count endpoints. However, as those are small changes, please feel free to merge afterwards, without returning for review.

PR #817, opened 2021-01-06
fix(rest): minor improvements to job endpoints

yT0n1:
Related to CAM-11543


yanavasileva (approved):
Looks good. Please adjust the example of response in set job retries async to contain the create user id.

yanavasileva (approved):

PR #821, opened 2021-01-11
chore(docs): Add documentation for added extension element + new API

mboskamp:
Related CAM-12040


tmetzke (commented):
Description looks good to me.
Should we add a link to the extension element in `reference/bpmn20/tasks/service-task.md#camunda-extensions`

PR #839, opened 2021-01-27
chore(rest): add cleaned up task metrics option to metrics

tmetzke:
related to CAM-13016


tasso94 (changes requested):
Please incorporate adjustments when changing to send date by path parameter.

PR #857, opened 2021-02-12
Fix typo in url port

timstawowski:
Just came across this typo while reading the docs.

### DIFF
--- a/content/reference/rest/deployment/post-redeploy-deployment.md
+++ b/content/reference/rest/deployment/post-redeploy-deployment.md
@@ -185,7 +185,7 @@ Status 200.
     "links": [
         {
             "method": "GET",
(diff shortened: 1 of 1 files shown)


koevskinikola:
Hi @timstawowski,

Thank you for the contribution. Why do you think this is a typo?

The link is provided as part of an example, and port `38080` is a completely valid port that can be mapped to the Camunda Platform Rest API. Furthermore, we use the same port in other examples of the Deployment endpoint, changing this one would make them inconsistent.

Best,
Nikola

timstawowski:
Hi @koevskinikola,

sure, `38080` indeed is a valid port. I was thinking you would stick to the `8080` port reference like in many other sections of the doc. This is why I thought that in this case, it would be a typo if that is obviously not the case, I will close this PR.

Best regards,
Tim

PR #859, opened 2021-02-19
feat(repository): display preview of HTML files

marstamm:
related to CAM-13104


tasso94 (changes requested):
Since the feature is activated by default, should we also consider documenting the possibility of unintended script execution in the migration guide? Maybe we can add a sentence about the impact in our [Security Instructions](https://docs.camunda.org/manual/latest/user-guide/security/) and link to it from the migration guide?

tasso94 (approved):

PR #866, opened 2021-02-26
[CAM-13161] chore(connect): correct path to ConnectorConfigurator

andue:
The file name is correct in the provided link `extending Connect` 

using version 7.14.0


tmetzke (approved):
Hi @andue,

thanks for spotting and correcting this.

Best,
Tobias

PR #875, opened 2021-03-10
Documentation fix

aloksingh25:


mboskamp:
Hi @aloksingh25,
thank you for contributing to the docs. However, I think this change is not valid.
We use Hugo for converting markdown to HTML and the links you changed will be adjusted by Hugo automatically.
In fact, this change breaks the Hugo conversion.
I will close this PR. However, feel free to comment and explain why you think this change should be included. Please also explain the current problem, that you are trying to fix.

Best,
Miklas

PR #896, opened 2021-03-22
Fix a section title in docs (external tasks).

matejd:
Just a small correction. Title was probably copied from another document.


tasso94:
Good catch! 👍 I merged it with our master.

PR #898, opened 2021-03-23
CAM-13303: fix(bpmn): adjust inclusive gateway limitations

koevskinikola:
* Provide a correct diagram of the limitation for inclusive gateways. The inclusive gateway will wait for multiple tokens, even though they will all arrive on the same sequence flow.

Related to CAM-13303


ThorbenLindhauer (changes requested):
Looks overall good, the descriptions are clear to follow. I suggest two improvements:

* Move the first example to the last place in the enumeration. It is the most complex example and an edge case
* In that example's description, we could explicitly say that according to the spec the gateway should trigger (most readers will not know what the spec defines for this case, so they will not know that the Camunda implementation violates the spec)

koevskinikola:
Implemented the hints. I also renamed the BPMN examples to reflect the order of the examples in the list.

ThorbenLindhauer (approved):
Perfectm, thank you!

PR #916, opened 2021-04-07
chore(screenshots): update screenshots to 7.15

tasso94:
related to CAM-13236


marstamm:
I have not looked through everything yet, but I noticed that all screenshots are double the resolution than the previous one, even when we explicitly resize the window: https://github.com/camunda/camunda-bpm-platform-ee/blob/master/webapps/camunda-webapp/plugins/src/test/js/screenshots/setup.js#L14

Most users don't profit from >Full HD pictures. Could this be caused by apple retina display?

tasso94:
I think it is because of the apple retina display. Do you think it is an issue? On most pages, only a few screenshots are displayed, leading to downloading a few hundred kilobytes more than before. Personally, especially when it comes to screenshots, I usually zoom into the image and prefer when it looks still crisp and not blurry.

marstamm:
I don;t think it's an issue, I was just surprised that this happens even when we explicitly set the resolution

tasso94:
We could also try to reduce the image size with something like https://www.npmjs.com/package/imagemin-pngquant 🤔

marstamm:
I had a slack time project to optimize the PNGs during build, but did not come around to create PR on the infra repo yet

marstamm (approved):
Looks good :rocket:

PR #917, opened 2021-04-07
chore(testing): update assert version to 10.0.0

yanavasileva:
Related to CAM-13258


tmetzke (approved):
Since the compatibility matrix has already been adjusted, this should cover all necessary parts 👍

PR #930, opened 2021-04-16
fix(docs): Prevent line break by shortening name

tobiasschaefer:
This fixes a glitch in the documentation, i.e. the long name "Micronaut Camunda Integration" previously forced a line break which does not look nice:

image

I fixed it by shorting the name. Note: I'm the author of the Micronaut Integration so I'm fine with the change :-)

Please also merge this to the 7.15 docs.


marstamm (approved):

marstamm:
Hi @tobiasschaefer ,

great catch! I'll go ahead and adjust the branches accordingly.

Cheers
Martin

marstamm:
We updated the page, you can check the result here: https://docs.camunda.org/manual/7.15/

tobiasschaefer:
thanks @marstamm , looks perfect :-)

PR #937, opened 2021-04-27
chore(rest): add missing parameter to historic incident endpoint

mboskamp:
Related to CAM-11562, CAM-10672

### DIFF
--- a/content/reference/rest/history/incident/get-incident-query.md
+++ b/content/reference/rest/history/incident/get-incident-query.md
@@ -115,6 +115,10 @@ GET `/history/incident`
     <td>Restricts to incidents that have the given parameter set as configuration.</td>
   </tr>
   <tr>
+    <td>historyConfiguration</td>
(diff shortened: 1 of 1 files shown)


content/reference/rest/history/incident/get-incident-query.md
```diff
@@ -114,6 +114,10 @@ GET `/history/incident`
     <td>configuration</td>
     <td>Restricts to incidents that have the given parameter set as configuration.</td>
   </tr>
+  <tr>
+    <td>historyConfiguration</td>
```
tmetzke, line 118: The attribute seems to be missing in the result and example as well (sorry for that, by the way 😊). Would you mind adding it there as well? 😃

tmetzke (changes requested):
Let's add the missing attribute consistently. 😃

tmetzke (approved):

PR #939, opened 2021-04-28
chore(rest): fix typo in historic batch

yT0n1:
Related to CAM-11556


tmetzke (approved):
Changes look all good 👍 
If there are further changes in https://github.com/camunda/camunda-bpm-platform/pull/1427, we should incorporate them here as well I guess.

PR #941, opened 2021-04-29
CAM-11569: chore(rest): add missing user op log properties

koevskinikola:
* Add missing User Operation log properties, and correct wrong property names.

Related to CAM-11569


yT0n1 (approved):
Looks good. Would you mind adding a type for `jobId` and `jobIdDefinition` while you are at it?

PR #942, opened 2021-04-30
CAM-11564: chore(rest): clarify cleanup endpoint limitation

koevskinikola:
* Make it clearer why the response of the `POST /history/cleanup` endpoint is unreliable from version 7.9.0.

Related to CAM-11564


tmetzke (approved):
I like the new explanation very much! 👍 
In case you find any suggestions in https://github.com/camunda/camunda-bpm-platform/pull/1432 useful, we should probably add them here as well. 😃

tmetzke (approved):

PR #958, opened 2021-05-17
docs(security): add note about java serialization format

tasso94:
related to CAM-13453


tmetzke (approved):
Nice read, I like the explicit addition 👍 
Should we also reference it in the warning at the bottom of https://docs.camunda.org/manual/latest/user-guide/process-engine/variables/#object-values? We could add something like "Please make sure to read the security implication prior to enabling this" or similar. What do you think? 🙂

tasso94:
Good point 👍

PR #991, opened 2021-06-11
docs(tasklist): make adding procInstId to task comment configurable

tasso94:
related to CAM-13609

### DIFF
--- a/content/webapps/tasklist/configuration.md
+++ b/content/webapps/tasklist/configuration.md
@@ -165,6 +165,21 @@ export default {
 
 **Note:** This does only affect the Tasklist login page. For other webapps, you need to adjust the corresponding config file as well.
 
+# Assign Process Instance Id to Task Comments
+
+When creating a Task Comment, the process instance ID is not assigned by default. 
+Queries for comments by process instance ID will not include those comments. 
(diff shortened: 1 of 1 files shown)


tasso94:
⚠️ Rebase after #989 ist merged!

content/webapps/tasklist/configuration.md
```diff
@@ -165,6 +165,21 @@ export default {
 
 **Note:** This does only affect the Tasklist login page. For other webapps, you need to adjust the corresponding config file as well.
 
+# Assign Process Instance Id to Task Comments
+
+By default, only the id of a task is assigned when creating a Task Comment with Tasklist. When you 
+want to query the created Task Comments later on by process instance id, this is impossible. When
```
marstamm, line 171: I think the first sentence does not convey the Problem clearly. I would reword it so that it's clearer that the process instance ID is missing.

```
When creating a Task Comment, the process instance ID is not assigned by default. 
Queries for comments by process instance ID will not include those comments.
```

tasso94: 👍

marstamm (approved):

PR #992, opened 2021-06-11
docs(rest): allow assigning a procInstId to a task comment

tasso94:
related to CAM-13610


yanavasileva (approved):
Looks good. Please have a look at the create comment request example. Already approving as it's small adjustment.

PR #1001, opened 2021-06-18
Update delete.md

StephenOTT:


yanavasileva:
Hi @StephenOTT,

Is there a reason for raising the PR in `7.15` branch, are there requirements to adjust the documentation for 7.15 version?
The contributions are raised for master branch (current development version) and backported by the team case by case if it's necessary. Otherwise, the contribution can be overlook and not pushed to master so the adjustment will not be carried for future versions of the docs. 
If the adjustment is not made on purpose for `7.15`, I will submitted it to `master` only and you won't need to create a new PR from master branch and so on.

Other that than, all look good, just a small suggestion for the wording that I will add below.

Best,
Yana

content/reference/rest/task/delete.md
```diff
@@ -12,7 +12,7 @@ menu:
 
 ---
 
-Removes a task by id.
+Removes a task by id.  Only tasks that are not part of a running process or case can be deleted / only custom tasks can be deleted.
```
yanavasileva, line 15: Let's replace `custom` with `standalone` as described here: https://docs.camunda.org/manual/develop/webapps/tasklist/dashboard/#create-a-standalone-task
```suggestion
Removes a task by id.  Only tasks that are not part of a running process or case can be deleted; only standalone tasks can be deleted.
```

yanavasileva (changes requested):

yanavasileva:
As I didn't receive any feedback, I adjusted the commit message, and merged the change to master with (66396a8).
@StephenOTT, thank you for your contribution.

PR #1028, opened 2021-07-16
chore(modeler): update installation to include forms / exclude cmmn

MaxTru:
closes https://jira.camunda.com/browse/CAM-13730


volkergersabeck (changes requested):
We should add a sentence to the '## Matching Camunda Process Engine Version' section that forms only work with Platform 7.15 and above

MaxTru:
> We should add a sentence to the '## Matching Camunda Process Engine Version' section that forms only work with Platform 7.15 and above

Good point, done

volkergersabeck (approved):
looks good to me now

PR #1036, opened 2021-08-10
chore(installation): clarify table prefix for Liquibase

tmetzke:
related to CAM-13799

### DIFF
--- a/content/installation/database-schema.md
+++ b/content/installation/database-schema.md
@@ -70,7 +70,7 @@ Consult the [manual installation guide]({{< ref "/installation/full/_index.md" >
 {{< note title="Heads Up!" class="info" >}}
 Liquibase provides additional commands to preview all changes that will be applied by commands that execute SQL statements on a database. For the `update` command, you can execute the [updateSql](https://docs.liquibase.com/commands/community/updatesql.html) command. This will let you inspect all statements that Liquibase will issue on your database without actually executing them.
 
-If you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts in the `liquibase/baseline` directory accordingly so that the tables are created with the prefix.
+Furthermore, if you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts in the `liquibase/baseline` directory accordingly so that the tables are created with the prefix.
 {{< /note >}}
 
 ## Manual installation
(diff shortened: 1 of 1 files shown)


content/installation/database-schema.md
```diff
@@ -67,8 +67,10 @@ As you create the tables externally via Liquibase, you have to configure the eng
 Set the `databaseSchemaUpdate` property to `false` (or, in case you are using Oracle, to `noop`).
 Consult the [manual installation guide]({{< ref "/installation/full/_index.md" >}}) of your distribution for further information on how to achieve that.
 
-{{< note title="Dry run" class="info" >}}
+{{< note title="Heads Up!" class="info" >}}
 Liquibase provides additional commands to preview all changes that will be applied by commands that execute SQL statements on a database. For the `update` command, you can execute the [updateSql](https://docs.liquibase.com/commands/community/updatesql.html) command. This will let you inspect all statements that Liquibase will issue on your database without actually executing them.
+
+If you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts in the `liquibase/baseline` directory accordingly so that the tables are created with the prefix.
```
tmetzke, line 73:
```suggestion
Furthermore, if you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts in the `liquibase/baseline` directory accordingly so that the tables are created with the prefix.
```

mboskamp:
There is another `Heads up!` box with a similar message. We could think about making them consistent. I cannot make suggestions for non-changed code, unfortunately.

tmetzke:
Could you explain what you would like to change? 🙂
I also don't like that you cannot make suggestions at unchanged code 👍

mboskamp:
In the current version of the file in lines 96 to 98 we have:

```
{{< note title="Heads Up!" class="info" >}}
If you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts accordingly so that the tables are created with the prefix.
{{< /note >}}
```

We could think about using the same wording there:
```
If you have defined a specific prefix for the entities of your database, you will have to manually adjust the `create` scripts in the `liquibase/baseline` directory accordingly so that the tables are created with the prefix.
```

Now looking at it more closely, maybe it's not worth the effort 😀

tmetzke:
OK, I see. I was actually thinking about naming the folder to adjust scripts at in the second note as well but then realized the create scripts are very well defined in that section already. Their location should be already clear to users. In the Liquibase section, we never mentioned those scripts specifically before so I thought it might make sense to make that more explicit.

mboskamp (approved):

PR #1044, opened 2021-08-23
docs(user-guide): document exclusive batch operation

tmetzke:
* describes how the set variables batch can be scheduled exclusively

related to CAM-13818


tasso94 (approved):
Looks good! 🙂 👍

PR #1047, opened 2021-08-23
CAM-13836: doc(quarkus): add quarkus deployments docs

koevskinikola:
Related to CAM-13836


tasso94 (changes requested):
Looks good to me. Just one suggestion to consider:
Can we change the page and sidebar title to `Resource Deployment`? Like this, it is easier to distinguish from the general quarkus application deployment.

tasso94 (approved):

PR #1063, opened 2021-09-01
CAM-13875: add paragraph for deployment criticality

ThorbenLindhauer:
- this should make it even clearer that users must ensure that attackers
  cannot perform deployments since Camunda does not implement a sandbox
  for BPMN execution or forms rendering

related to CAM-13875


tasso94 (approved):
Looks good to me!

PR #1064, opened 2021-09-01
docs(rest): organizing layout of API parameters with possible values/…

christinaausley:
…defaults

Related to CAM-13848


christinaausley:
In many of our docs, including [these](https://docs.camunda.org/manual/7.15/reference/rest/overview/), we do refer to `type` as `value`. Perhaps we could consider adjusting these, either in this branch or a new one? Open to thoughts/feedback on this change.

PR #1069, opened 2021-09-02
feat(job-executor): add documentation for job executor prio range

mboskamp:
Related to CAM-13486


content/user-guide/process-engine/the-job-executor.md
```diff
@@ -364,6 +366,16 @@ For example:
   </tr>
 </table>
 
+## Job Executor priority range
+
+By default, the Job Executor executes all jobs regardless of their priorities. Some jobs might be more important to finish quickly than others, so we assign them priorities and set `jobExecutorAcquireByPriority` to `true` as described above. Depending on the work load the Job Executor might be able to execute all jobs eventually. But if the load is high enough, we can face potential blocks where a Job Executor is always busy working on the high priority jobs and never manages to execute the lower priority jobs.
```
tasso94, line 371:
```suggestion
By default, the Job Executor executes all jobs regardless of their priorities. Some jobs might be more important to finish quicker than others, so we assign them priorities and set `jobExecutorAcquireByPriority` to `true` as described above. Depending on the workload, the Job Executor might be able to execute all jobs eventually. But if the load is high enough, we might face starvation where a Job Executor is always busy working on high-priority jobs and never manages to execute the lower priority jobs.
```

Isn't starvation the thing here?

content/user-guide/process-engine/the-job-executor.md
```diff
@@ -364,6 +366,16 @@ For example:
...
+
+To prevent this, you can specify a priority range for the job executor by setting values for `jobExecutorPriorityRangeMin` or `jobExecutorPriorityRangeMax` (or both). The Job Executor will only acquire jobs that are inside its priority range (inclusive). Both properties are optional, so it is fine to only set one of them.
```
tasso94, line 373:
```suggestion
To prevent this, you can specify a priority range for the job executor by setting values for `jobExecutorPriorityRangeMin` or `jobExecutorPriorityRangeMax` (or both). The Job Executor will only acquire jobs that are inside its priority range (inclusive). Both properties are optional, so it is fine only to set one of them.
```

content/user-guide/process-engine/the-job-executor.md
```diff
@@ -364,6 +366,16 @@ For example:
...
+
+To avoid job starvation, make sure to have no gaps between Job Executor priority ranges. If, for example, Job Executor A has a priority range of 0 to 100 and Job Executor B executes jobs from priority 200 to `Long.MAX_VALUE` any job that receives a priority of 101 to 199 will never be executed. Job starvation can also occur with `batch jobs` and `history cleanup jobs` as both types of jobs also receive priorities (default: `0`). You can configure them via their respective properties: `batchJobPriority` and `historyCleanupJobPriority`.
+
+This feature is particularly useful if you want to separate multiple types of jobs from each other. For example short-running, urgent jobs with high priority and long-running jobs that are not urgent but should finish eventually. Setting up a Job Executor priority range for both types will ensure that the long-running jobs can not block the urgent jobs.
```
tasso94, line 377:
```suggestion
This feature is particularly useful if you want to separate multiple types of jobs from each other. For example, short-running, urgent jobs with high priority and long-running jobs that are not urgent but should finish eventually. Setting up a Job Executor priority range for both types will ensure that long-running jobs can not block urgent ones.
```

content/user-guide/process-engine/the-job-executor.md
tasso94: Can we add a direct link to jump to the right place on the process engine config page?
E.g., to `process-engine.md#jobExecutorPriorityRangeMin` and `process-engine.md#jobExecutorPriorityRangeMax`

content/reference/deployment-descriptors/tags/process-engine.md
```diff
@@ -617,6 +617,31 @@ The following is a list with the most commonly used process engine configuration
         <strong>Values:</strong> <code>true</code>, <code>false</code> (Boolean).
       </p>
   </td>
+
+  <tr>
```
tasso94, line 621:
```suggestion
  <tr id="jobExecutorPriorityRangeMin">
```
Like this, we can add a direct link to this property.

content/reference/deployment-descriptors/tags/process-engine.md
```diff
@@ -617,6 +617,31 @@ The following is a list with the most commonly used process engine configuration
...
+    <td>
+        When set, the job executor will only acquire jobs that have a priority equal to or higher than the specified threshold.
+        You can combine this property with <code>jobExecutorPriorityRangeMax</code> to specify a job priority range for the job executor.
+        Note, that you can configure the priority of batch jobs and history cleanup jobs via the <code>batchJobPriority</code> and <code>historyCleanupJobPriority</code> properties. Both default to <code>0</code>.
+      <p>
+        <strong>Default value:</strong> <code>null</code>
+      </p>
+  </td>
+
+  <tr>
```
tasso94, line 633:
```suggestion
  <tr id="jobExecutorPriorityRangeMax">
```
Like this, we can add a direct link to this property.

tasso94 (approved):
Looks good to me. Just some smaller ideas/hints you can find below ⬇️

PR #1070, opened 2021-09-03
docs(user-guide): add CorrelateMessage batch operation

tmetzke:
* adds the CorrelateMessage batch operation to the list of operations
* adds the user operation log entry created by the CorrelateMessage batch

related to CAM-13863, CAM-13819

### DIFF
--- a/content/user-guide/process-engine/batch-operations.md
+++ b/content/user-guide/process-engine/batch-operations.md
@@ -24,6 +24,7 @@ The following operations can be executed asynchronously
 - [Set a Removal Time to Historic Process Instances](#historic-process-instances)
 - [Set a Removal Time to Historic Decision Instances](#historic-decision-instances)
 - [Set a Removal Time to Historic Batches](#historic-batches)
+- [Correlate Messages to Process Instances](#correlate-messages-to-process-instances)
 
 All batch operations rely on corresponding methods that provide the possibility to
 operate on a list of entities synchronously. Please refer to the general [Batch][batch] documentation to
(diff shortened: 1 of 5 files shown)


tasso94 (changes requested):
Looks good to me in general. 

Just some hints at which other places in the docs we could mention the new batch operation as well:
* Should we add the new batch operation to the configuration map `batchOperationsForHistoryCleanup` [example](https://docs.camunda.org/manual/develop/user-guide/process-engine/history/#batches)?
* Should we link back to your documentation by adding the batch operation to [this list](https://docs.camunda.org/manual/develop/user-guide/process-engine/batch/#creating-a-batch)?
* Should we reflect the "create batch" permission [here](https://docs.camunda.org/manual/develop/user-guide/process-engine/authorization-service/#additional-batch-permissions)?

WDYT?

tmetzke:
> Should we add the new batch operation to the configuration map batchOperationsForHistoryCleanup example?

Yes, very good idea!  👍 
I assume we need to adjust the backend as well then to take it into consideration?

> Should we link back to your documentation by adding the batch operation to this list?

I didn't have a look at all the links in both lists (the one on the "Batch" page and the one on the "Batch operations" page), but how about we remove the list from the "Batch" page and reference to the "Batch operations" page instead? Is there a specific need to maintain the list twice?

> Should we reflect the "create batch" permission here?

Yes, definitely! 👍

tasso94:
> Is there a specific need to maintain the list twice?

Fully agree with your proposal to reduce maintenance efforts.

> I assume we need to adjust the backend as well then to take it into consideration?

Should work out of the box.

tmetzke:
>> I assume we need to adjust the backend as well then to take it into consideration?

> Should work out of the box.

OK, found the code for that, very nice! 👍

content/user-guide/process-engine/batch-operations.md
```diff
@@ -21,20 +21,26 @@ The following operations can be executed asynchronously
...
-In particular, this means that the seed job and execution jobs will receive a `deploymentId` so [deployment-aware job executors]({{< ref "/user-guide/process-engine/the-job-executor.md#job-execution-in-heterogeneous-clusters" >}}) can pick up those jobs of a batch that need to be executed on their nodes. The deployment id of the seed job is chosen from the list of involved deployments. This list is derived from the resulting set of affected instances. Execution jobs only contain elements of the same deployment and are bound to that deployment's id.
+In particular, this means that the seed job and execution jobs will receive a `deploymentId` so 
+[deployment-aware job executors]({{< ref "/user-guide/process-engine/the-job-executor.md#job-execution-in-heterogeneous-clusters" >}}) 
+can pick up those jobs of a batch that need to be executed on their nodes. 
+The deployment id of the seed job is chosen from the list of involved deployments. 
+This list is derived from the resulting set of affected instances. 
+Execution jobs only contain elements of the same deployment and are bound to that deployment's id.
```
tmetzke, line 43: I didn't adjust anything here content-wise, just added newlines

tasso94:
One more thing: should we link back to your documentation from the [Message API](https://docs.camunda.org/manual/develop/reference/bpmn20/events/message-events/#message-api) documentation?

tmetzke:
I thought about it but couldn't find a good place to add it, actually. Since it also works differently than the synchronous correlation (no message start events, other parameters, different filtering), I decided against it for now. 

Let me know if you think otherwise 🙂

tasso94 (approved):
Nice 🚀 💯

PR #1075, opened 2021-09-13
Update manual-task.md

fml2:


koevskinikola (approved):
Nice catch. Thanks for fixing it!

PR #1077, opened 2021-09-16
feat(was-install): improve variable explanation

falko:
Mention that variables don't have to be created and that version numbers can contain the `-ee` suffix (or always will for the EE-only WAS distro).

This is based on feedback from a prospect that was uncertain about whether the right files have been downloaded.


tasso94:
Thanks, @falko, makes a lot of sense!

PR #1084, opened 2021-09-24
chore(database): update erd to 7.16

tmetzke:
related to CAM-13924


tasso94 (changes requested):
Looks good to me in general. Just two minor hints/questions:
* Can we move the form table in the upper left corner to the other tables that don't have relations? 
* The jobdef table is slightly cut-off by the canvas. Can we move it a little bit up or increase the size of the canvas?

engine-erd

tmetzke:
Sure, good hints 👍

tasso94 (approved):

PR #1089, opened 2021-10-05
chore(release): update compatibility matrices

yT0n1:
Related to CAM-13934, CAM-13933


tmetzke (approved):
New versions look good 👍

PR #1129, opened 2021-12-13
docs(rest-api): update task claim and task assignee

epollum:
- Update docs to reflect that if the userId is empty the task is unclaimed.

related to CAM-14071


tmetzke (approved):
Looks good to me 👍

PR #1148, opened 2022-02-02
CAM-13338: doc(run): add guide on how to add process engine plugins

koevskinikola:
Related to CAM-13338

### DIFF
--- a/content/user-guide/camunda-bpm-run.md
+++ b/content/user-guide/camunda-bpm-run.md
@@ -353,11 +353,18 @@ camunda.bpm.run.process-engine-plugins:
     plugin-parameters:
       parameterOne: valueOne
       parameterTwo: true
-  - 
 ```
 
 In the example above, we use the `TestPlugin` canonical name as a YAML key. The YAML value consists of a 
 collection of key-value pairs that represent the configuration parameters for the `TestPlugin` and their values.
+Some process engine plugins don't have configuration parameters. For them, you only need to define the `plugin-class` 
+YAML property, like so:
+
+```yaml
+camunda.bpm.run.process-engine-plugins:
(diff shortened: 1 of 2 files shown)


content/user-guide/camunda-bpm-run.md
```diff
@@ -275,6 +275,75 @@ In the table below, observe the Camunda Run-specific properties for the Administ
...
+      <td><code>.process-engine-plugins</code></td>
+      <td>Define your process engine plugin configurations under this YAML property.</td>
+      <td>none</td>
+  </tr>
+</table>
+
+In Camunda Run, process engine plugins can be registered by performing the following steps:
+
+1. Always find and read the process engine plugin documentation.
+  * Find the canonical name of the process engine plugin Java class that implements the `ProcessEnginePlugin` interface.
```
mboskamp, line 307: To non-Java users, it might not be clear how to identify that class. They might not know what a Java interface is. We should educate our plugin creators on how to provide the information so that Run users can understand them. Might be a nice idea to talk to dev-rel on that too.

mboskamp (changes requested):
Very nice documentation!

I added a few comments regarding examples for plugins that can not be configured.
I think it would be great to have some resources for plugin creators on how to document their plugins for Run users. We have to keep in mind that they are not Java users and might not know how to identify an interface, for example.

koevskinikola:
@christinaausley the new docs that I am adding are intended for non-Java developers. If you have any additional hints regarding that type of documentation, please let me know. :slightly_smiling_face:

christinaausley (approved):
This is excellent. I made a few minor changes in my latest commit. I really liked how you outlined the first portion in steps, so I applied the same to the second portion of action items.

PR #1158, opened 2022-02-25
chore(theme): update docs theme

akeller:
Related to [CAM-14373](https://jira.camunda.com/browse/CAM-14373).


akeller:
@ThorbenLindhauer did I do something wrong? I wasn't expecting to see such a large amount of insertions 😅

ThorbenLindhauer (changes requested):
Hey, the changes you made look good. Two points from your goals are missing though:

> Main Page title - The Camunda Platform Manual -> The Camunda Platform 7 Manual

I think this you can fix here: https://github.com/camunda/camunda-docs-manual/blob/master/layouts/index.html

> Side-menu version - Camunda Platform -> Camunda Platform 7

I assume this is the header menu for switching between platform, Cawemo, Optimize, etc. In that case, this is controlled via the sections configuration here: https://github.com/camunda/camunda-docs-manual/blob/master/config.yaml

Re the large diff: You are replacing the minified versions of the CSS and JS with the non-minified ones. I don't mind if we publish the non-minified theme, I guess it won't make too big of a difference.

If you want to use the minified version, you can:

* First merge the docs-theme PR, which will trigger a Jenkins job that will publish the minified theme to the dist branch: https://github.com/camunda/camunda-docs-theme/tree/dist. You can then copy that into the consuming repo
* Or instead, run `grunt build` and then `grunt optimize`, and then manually replace the css and js from the theme with where you consume it

Please don't ask me why this is so complicated, I don't know 😅 

Lastly, if you want this to be applied to 7.16, 7.15, and 7.14, you would also have to raise PRs against those branches.

ThorbenLindhauer:
Re minified vs non-minified js and css: Since you only changed html and images, I believe you can also simply remove the changes to the CSS and JS from your commit. There should be no functional changes anyway.

akeller:
> Please don't ask me why this is so complicated, I don't know 😅

It's like you where my brain was going.

Let me take another shot at this.

akeller:
I think this is good now 🤞 I just got confused with what I was changing where and what files were in what repos 😅 

If this looks good, I'll raise PRs against those other branches as well.

ThorbenLindhauer (approved):
Looks good. I made one more change to add the 7 to the header menu: https://github.com/camunda/camunda-docs-manual/pull/1158/commits/cd63a6553c23c6fba954dec6a961fff2a5216c5e

Please revert if you don't need it. Else we can go ahead with squash and merge.

akeller:
Good catch! Thank you for your help here! I'll get the other PRs together for the branches.

PR #1166, opened 2022-03-02
docs(update): add latest camunda-bpm-assert version

danielkelemen:
related to CAM-14414


danielkelemen:
We won't document this release as it's no different than the previous one (it only contains internal changes).

PR #1172, opened 2022-03-10
Update post-start-process-instance.md

jyot-prakash:
The previous sentence was not making sense.


tmetzke:
Hi @jyot-prakash,

as far as I can see, the new sentence does not make much sense while the old one does.
It is the case instance id that the process instance is initialized with.
Thus, I will close this here and won't fix.

PR #1178, opened 2022-03-14
CAM-14140: deprecate initial data report

ThorbenLindhauer:
- moves the section to the end of the page because it is not relevant
  moving forward and only remains for users of previously released
  versions

related to CAM-14140


tmetzke (approved):
I like it 👍

PR #1184, opened 2022-03-24
CAM-14347: remove environments retired with 7.17

ThorbenLindhauer:
related to CAM-14347


tasso94 (commented):
What is the reason for keeping the documentation page `/user-guide/process-engine/database/mariadb-galera-configuration.md` when we drop support for Maria DB Galera Cluster? Looks good otherwise.

ThorbenLindhauer:
The only reason is that I put this into the scope of https://jira.camunda.com/browse/CAM-14266 :)

PR #1187, opened 2022-03-29
Change to the new artifactory domain

chaima-mnsr:
Related to INFRA-3183

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,6 @@ _Note: For screenshot of the enterprise webapps, you can use the automatic scree
 
 [Hugo]: http://gohugo.io/
 [Hugo Installation]: http://gohugo.io/overview/installing/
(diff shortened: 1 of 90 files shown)


Langleu (approved):
checked that the URL structure is correct but did not check that every single link is redirecting to something working but would assume everything is fine for the things that have been released already as only the domain changed.

Langleu:
Don't forget to backport the change till 7.4 I think as 7.0-7.3 had a completely different structure.

ThorbenLindhauer (changes requested):
Changes look good overall. Similar to https://github.com/camunda/camunda-docs-static/pull/257, there seem to be issues with links that point to the `internal` and `camunda-bpm-ee` groups. Let's try to resolve them first.

I will review the backports once we are finished with the master PR.

ThorbenLindhauer (approved):
As Chaima shared on slack, this is currently expected and will resolve once we communicate this to our users, so I will go ahead with merging.

PR #1203, opened 2022-04-07
CAM-14460: correct section about assert parent id change

ThorbenLindhauer:
- removes some more content that is unnecessary for an update guide

related to CAM-14460


tasso94 (approved):
Nice catch!

PR #1207, opened 2022-04-08
improve(setup): prescribe how to install old version of hugo

pepopowitz:
I ran into some issues while trying to install hugo v0.54 with homebrew, and @akeller suggested I install via tarball instead. That was much easier (and actually worked), so this PR: 

* Updates the setup docs to prescribe installing Hugo via tarball.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ Have a look at our [contribution guide](https://github.com/camunda/camunda-bpm-p
 
 ## Installing Hugo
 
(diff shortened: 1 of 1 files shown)


pepopowitz:
ah, do you mean in this line? 

> you first need to install [Hugo][Hugo] v0.54

Assuming that's what you meant, yes, it makes sense to me -- I will add another commit.

akeller (approved):
Great add! Thank you! 🚀

PR #1247, opened 2022-06-02
docs(process-engine): config property for bpmn stacktrace log level

danielkelemen:
related to CAM-13977

### DIFF
--- a/content/update/minor/717-to-718/_index.md
+++ b/content/update/minor/717-to-718/_index.md
@@ -21,7 +21,8 @@ This document guides you through the update from Camunda Platform `7.17.x` to `7
 4. For administrators and developers: [Groovy version update](#groovy-version-update)
 5. For administrators: [Camunda Docker Images: Base image updated to Alpine 3.15](#camunda-docker-images-base-image-updated-to-alpine-3-15)
 6. For administrators and developers: [XLTS for AngularJS](#xlts-for-angularjs)
-1. For administrators and developers: [Stricter default Content Security Policy](#stricter-default-content-security-policy)
+7. For administrators and developers: [Stricter default Content Security Policy](#stricter-default-content-security-policy)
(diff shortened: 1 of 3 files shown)


content/update/minor/717-to-718/_index.md
```diff
@@ -126,3 +127,9 @@ If you have added custom script tags in one of the `index.html` files of the web
 You don't need to worry about whitelisting for scripts you load via our plugin system.
 
 Find the details in the [Content Security Policy]({{< ref "/webapps/shared-options/header-security.md#content-security-policy" >}}) section.
+
+
+# Log level configuration for BPMN stack trace
+
+We've added a new configuration property called `logLevelBpmnStackTrace` to change the default `DEBUG` level of the bpmn stack trace.
+See the [Logging level parameters]({{< ref "/reference/deployment-descriptors/tags/process-engine.md#logLevelBpmnStackTrace" >}}) section for details.
```
koevskinikola, line 135: Should we also describe what the impact is? For example, tell users that the previous default remains and that they don't need to do any adjustments.

koevskinikola (changes requested):
Looks good. I only have two suggestions:
1. Should we reference the property in the logging section here: https://github.com/camunda/camunda-docs-manual/blob/master/content/user-guide/logging.md#logging-categories?
2. Should we expand the update section to explain to users what they need or needn't do for this change?

content/update/minor/717-to-718/_index.md
danielkelemen: Good points!
I added some extra infos about the impact and what they could do.
I also referenced the new parameters in the logging user-guide.
Let me know what you think!

koevskinikola (approved):

content/update/minor/717-to-718/_index.md
koevskinikola: Looks great!

PR #1259, opened 2022-06-30
fix(seo): update theme to incorporate semantic nav element

pepopowitz:
Updates the theme to incorporate https://github.com/camunda/camunda-docs-theme/pull/29. 

I intend on merging that one first, but I don't think it matters 🤷


akeller (approved):
I did not run this locally, but I'm confident you spelled nav correclty.

PR #1265, opened 2022-07-06
docs(user-guide/rest): document exception error code feature 

tasso94:
This documentation is related to the feature implemented in https://github.com/camunda/camunda-bpm-platform/pull/1943.

related to CAM-14083


christinaausley (approved):
Nice work here. No adjustments on my end for this one.

PR #1266, opened 2022-07-12
docs(batch): add batch startTime to API

danielkelemen:
related to CAM-14721

### DIFF
--- a/content/reference/rest/batch/get-query.md
+++ b/content/reference/rest/batch/get-query.md
@@ -138,7 +138,8 @@ Status 200.
     "batchJobDefinitionId": "aBatchJobDefinitionId",
     "suspended": false,
     "tenantId": "aTenantId",
-    "createUserId": "aUserId"
+    "createUserId": "aUserId",
+    "startTime":"2013-01-23T13:42:42.000+0200"
   }
 ]
 ```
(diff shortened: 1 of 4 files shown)


content/reference/rest/batch/get-query.md
```diff
@@ -62,7 +62,7 @@ GET `/batch`
     <td>sortBy</td>
     <td>
       Sort the results lexicographically by a given criterion.
-      Valid values are <code>batchId</code> and <code>tenantId</code>.
+      Valid values are <code>batchId</code>, <code>tenantId</code> and <code>startTime</code>.
```
ThorbenLindhauer, line 65: This and the result property should go to `get-statistics-query.md`.

ThorbenLindhauer (changes requested):
It looks like you updated the wrong endpoint description, please check the comment.

content/reference/rest/batch/get-query.md
danielkelemen: You're right, the new sortBy property is only available for the statistics query.
What did you mean by the result property? The new startTime field is returned for both batch and statistics endpoints.

ThorbenLindhauer: Yeah, you are right. I don't recall why I wrote it like that.

ThorbenLindhauer (approved):
Looks good!

PR #1309, opened 2022-09-27
chore(was): improve liberty installation guide

mboskamp:
Related to CAM-14913


koevskinikola (approved):
All good. Nice catch with the EAR location!

PR #1318, opened 2022-10-06
docs(mig-guide): discontinuation of `engine-rest-jaxrs2` artifact

tasso94:
related to CAM-14923

### DIFF
--- a/content/update/minor/717-to-718/_index.md
+++ b/content/update/minor/717-to-718/_index.md
@@ -27,6 +27,7 @@ This document guides you through the update from Camunda Platform `7.17.x` to `7
 1. For developers: [Adjusted exception handling in the Java External Task Client](#adjusted-exception-handling-in-the-java-external-task-client)
 1. For administrators: [Batch execution start time](#batch-execution-start-time)
 1. For developers: [Discontinue Camunda H2 console webapp](#discontinue-camunda-h2-console-web-app)
+1. For administrators and developers: [REST API artifact `camunda-engine-rest-jaxrs2` discontinued](#rest-api-artifact-camunda-engine-rest-jaxrs2-discontinued)
 
 This guide covers mandatory migration steps and optional considerations for the initial configuration of new functionality included in Camunda Platform 7.18.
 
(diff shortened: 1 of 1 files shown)


content/update/minor/717-to-718/_index.md
```diff
@@ -208,3 +209,9 @@ Including it in a ready-to-use distribution means an additional step for end use
 
 Going forward we will not further develop the Camunda H2 console app. This also means there will not be any more releases for this project.
 If you need to connect to the default file-based H2 database during development, we encourage you to use third-party tools like [DBeaver](https://dbeaver.io/).
+
+# REST API artifact `camunda-engine-rest-jaxrs2` discontinued
+
+Starting with 7.18.0, we discontinue the REST API artifact `org.camunda.bpm:camunda-engine-rest-jaxrs2`. We provided this artifact to allow long-polling on the REST API endpoint `/external-task/fetchAndLock` for application servers/runtimes that support the JAX-RS 2.0 specification. Application servers/runtimes only supporting JAX-RS 1.0 had to use the artifact `org.camunda.bpm:camunda-engine-rest-core`, excluding the long-polling capability. 
+
+Since we dropped support for IBM WebSphere 8.5 with the 7.17.0 release, we could consolidate both artifacts again into the `org.camunda.bpm:camunda-engine-rest-core` artifact, which is now based on the JAX-RS 2.0 specification and provides support for long-polling.
```
tasso94, line 217:
```suggestion
Since we dropped support for IBM WebSphere 8.5 with the 7.17.0 release, we were able to consolidate both artifacts again into the `org.camunda.bpm:camunda-engine-rest-core` artifact, which is now based on the JAX-RS 2.0 specification and provides support for long-polling.
```
Hey @christinaausley,

Miklas and I weren't sure whether to choose `could` or `were able to`. Grammarly prefers `could`. What do you think?

akeller: 👋 Chiming in because @christinaausley is on vacation. 

I would recommend `can` or `could` because we are talking about the action we can do (or did) now.

I tried to find a good resource for this with examples if you and @mboskamp wanted a more in-depth explanation.

* https://english.stackexchange.com/questions/122310/was-able-to-vs-could
* https://blogs.transparent.com/english/cancould-vs-to-be-able-to/ - this has the best explanation for `can` vs. `could` IMO.

Still, I would love to hear from @christinaausley on this topic. It's a good one! But I also want to make sure you don't feel blocked.

mboskamp: @akeller, we already merged the change (including `could` instead of `were able to`) so we are not blocked. This PR is here to possibly improve the docs and understand what the best wording is.
My thinking was, that using `could` might sound like we are talking about a potential thing that we `could` do whereas using `we were able to` it is clear that we already performed this action. Maybe this is just my German brain talking.

christinaausley: All of the conversation here is great and makes sense -- it's sounding largely like a past/present/future action determination.

If this change has already taken place, I would take out could/can/were able to to avoid confusion, and just simply say "Since we X... **we consolidated** both artifacts again into...".

If this change has not yet taken place, I would say "Since we X... **we plan to consolidate** both artifacts again into...".

tasso94: Thanks, @christinaausley, I went for removing any uncertainty and changed the verb to "consolidated".

christinaausley (approved):

PR #1325, opened 2022-10-24
chore(rest): improve REST docs

mboskamp:
Related to camunda/camunda-bpm-platform#2853

### DIFF
--- a/content/reference/rest/history/process-instance/get-duration-report.md
+++ b/content/reference/rest/history/process-instance/get-duration-report.md
@@ -8,7 +8,7 @@ menu:
     name: "Get Duration Report"
     identifier: "rest-api-history-get-duration-report"
     parent: "rest-api-history-process-instance"
-    pre: "GET `/history/process-instance/report?reportType=duration&periodUnit=month`"
+    pre: "GET `/history/process-instance/report`"
 
 ---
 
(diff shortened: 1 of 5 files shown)


content/reference/rest/history/process-instance/get-duration-report.md
```diff
@@ -8,7 +8,7 @@ menu:
     name: "Get Duration Report"
     identifier: "rest-api-history-get-duration-report"
     parent: "rest-api-history-process-instance"
-    pre: "GET `/history/process-instance/report?reportType=duration&periodUnit=month`"
```
danielkelemen, line 11: I guess there are no other report types right now, only duration, right?

danielkelemen (approved):
Looks good!

content/reference/rest/history/process-instance/get-duration-report.md
mboskamp: Correct, right now only `duration` [is supported](https://github.com/camunda/camunda-bpm-platform/blob/ee11d92faeb0f3e504098b0ce378de84a2ae186b/engine-rest/engine-rest/src/main/java/org/camunda/bpm/engine/rest/dto/history/HistoricProcessInstanceReportDto.java#L50).
But either way, we never put the query params in the method description even if they are mandatory. This is the only place we did it.

PR #1327, opened 2022-11-03
docs(history): improve documentation for history Removal-Time Strategy

Nishant-sehgal:
related to camunda/camunda-bpm-platform/issues/2829

### DIFF
--- a/content/user-guide/process-engine/history.md
+++ b/content/user-guide/process-engine/history.md
@@ -1904,7 +1904,7 @@ Removal time is defined per instance as `removal time = base time + TTL`. `base 
 
 Valid values are `start`, `end` and `none`. `end` is the default value and the recommended option. `start` is a bit more efficient when the process engine populates the history tables, because it does not have to make extra `UPDATE` statements when an instance finishes.
 
-**Note:**: With `historyRemovalTimeStrategy` set to `start`, it is possible to delete historic data of running process instances.So whenever a process is started removal time will be calculated(start+ttl) and will be set for all the activities of the process.As soon as removal time is reached data from historic tables gets cleaned up irrespective of process is running or completed.This may lead to removal of the historic data before the process instance is finished resulting into no available history in cockpit or history tables.
+**Note:**: With `historyRemovalTimeStrategy` set to `start`, it is possible to delete historic data of running process instances. When a process is started the removal time will be calculated (start+TTL) and will be set for all the activities of the process. As soon as removal time is reached data from historic tables gets cleaned up irrespective of whether the instance is running or completed. This may lead to the removal of the historic data before the process instance is finished resulting in no available history in Cockpit or history tables. A mitigation strategy is to choose a longer TTL value or set `historyRemovalTimeStrategy` to `end`.
 
 {{< note title="Heads-up!" class="info" >}}
 The calculation of the removal time can be enabled independently of the selected cleanup strategy of the process engine.
--- a/content/user-guide/process-engine/history.md
+++ b/content/user-guide/process-engine/history.md
@@ -1904,6 +1904,8 @@ Removal time is defined per instance as `removal time = base time + TTL`. `base 
 
 Valid values are `start`, `end` and `none`. `end` is the default value and the recommended option. `start` is a bit more efficient when the process engine populates the history tables, because it does not have to make extra `UPDATE` statements when an instance finishes.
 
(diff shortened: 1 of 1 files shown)


content/user-guide/process-engine/history.md
```diff
@@ -1903,6 +1903,7 @@ Removal time is defined per instance as `removal time = base time + TTL`. `base
 ```
 
 Valid values are `start`, `end` and `none`. `end` is the default value and the recommended option. `start` is a bit more efficient when the process engine populates the history tables, because it does not have to make extra `UPDATE` statements when an instance finishes.
+Note: With "historyRemovalTimeStrategy" set to "start", it is possible to delete historic data of running process instances.
```
mboskamp, line 1906:
```suggestion
Note: With `historyRemovalTimeStrategy` set to `start`, it is possible to delete historic data of running process instances.
```

mboskamp (changes requested):
Hi @Nishant-sehgal,
Thank you very much for picking up camunda/camunda-bpm-platform#2829 and working on it!
I think it would be good to give a bit more background information. The user wants to know what to look out for when using `historyRemovalTimeStrategy=start`.

Here is a bit of information from the ticket where this issue was brought up. Maybe you can use it to make the added documentation even more helpful:

* When you access the Cleanup Console, you can see a list of versioned process definitions but only if they have, at least, one finished instance. If no instance is finished, for a given process definition (version), you will not see anything about this process definition/version. So I thought that only finished instances are cleanable.
* If you perform a batch operation setting the removal time, without specifying to operate only on finished instances, you will modify that time even to unfinished instances, and you may have an odd situation in which the cleanable instances are more than the finished ones.
* If you then launch the Clean Up process, all the cleanable instances will be wiped from the History Tables (finished or not).
* So, when someone works on the unfinished instances, and they finish, you will never have a History for them (they seem to go on without any error)

Nishant-sehgal:
> Hi @Nishant-sehgal, Thank you very much for picking up [camunda/camunda-bpm-platform#2829](https://github.com/camunda/camunda-bpm-platform/issues/2829) and working on it! I think it would be good to give a bit more background information. The user wants to know what to look out for when using `historyRemovalTimeStrategy=start`.
> 
> Here is a bit of information from the ticket where this issue was brought up. Maybe you can use it to make the added documentation even more helpful:
> 
>     * When you access the Cleanup Console, you can see a list of versioned process definitions but only if they have, at least, one finished instance. If no instance is finished, for a given process definition (version), you will not see anything about this process definition/version. So I thought that only finished instances are cleanable.
> 
>     * If you perform a batch operation setting the removal time, without specifying to operate only on finished instances, you will modify that time even to unfinished instances, and you may have an odd situation in which the cleanable instances are more than the finished ones.
> 
>     * If you then launch the Clean Up process, all the cleanable instances will be wiped from the History Tables (finished or not).
> 
>     * So, when someone works on the unfinished instances, and they finish, you will never have a History for them (they seem to go on without any error)

Thanks for sharing the details.

The details you shared are specifically if some one use the cockpit to set the removal time which is different i believe because there is no historyRemovalTimeStrategy coming in picture.

What i have understood so far is `historyRemovalTimeStrategy= start` means whenever a process is started removal time will be calculated(start+ttl) and will be set for the involved entities.As soon as removal time is reached it gets cleaned up irrespective of process is running or finished.

Let me know shall i add more details or the info that is added is enough.

Nishant-sehgal:
@mboskamp Can you check if it looks good ?

content/user-guide/process-engine/history.md
```diff
@@ -1904,6 +1904,8 @@ Removal time is defined per instance as `removal time = base time + TTL`. `base
 
 Valid values are `start`, `end` and `none`. `end` is the default value and the recommended option. `start` is a bit more efficient when the process engine populates the history tables, because it does not have to make extra `UPDATE` statements when an instance finishes.
 
+**Note:**: With `historyRemovalTimeStrategy` set to `start`, it is possible to delete historic data of running process instances.So whenever a process is started removal time will be calculated(start+ttl) and will be set for all the activities of the process.As soon as removal time is reached data from historic tables gets cleaned up irrespective of process is running or completed.This may lead to removal of the historic data before the process instance is finished resulting into no available history in cockpit or history tables.
```
mboskamp, line 1907:
```suggestion
**Note:**: With `historyRemovalTimeStrategy` set to `start`, it is possible to delete historic data of running process instances. When a process is started the removal time will be calculated (start+TTL) and will be set for all the activities of the process. As soon as removal time is reached data from historic tables gets cleaned up irrespective of whether the instance is running or completed. This may lead to the removal of the historic data before the process instance is finished resulting in no available history in Cockpit or history tables. A mitigation strategy is to choose a longer TTL value or set `historyRemovalTimeStrategy` to `end`.
```

mboskamp (changes requested):
Hi @Nishant-sehgal, 
Thank you for going over the paragraph again. Now it is much better in my opinion.
I added some suggestions on how we can further improve the section. Could you please have a look and let me know if the changes work for you?
Cheers,
Miklas

mboskamp (approved):

PR #1333, opened 2022-11-07
feat: canonicals(master) -- correctly fix sitemap & apply rel=canonical strategy to master branch

pepopowitz:
Part of https://github.com/camunda/product-hub/issues/232.

1. Updates the theme based on https://github.com/camunda/camunda-docs-theme/pull/33
2. Specifies a `canonicalBasePath` setting for proper sitemap & rel=canonical generation
 
## Proof that the sitemap gets generated with absolute URLs

I ran a build locally and this is the generated sitemap:

image

## Proof that pages are generated with a versionless rel=canonical

I ran a build locally and this is the generated page for `/manual/develop/user-guide/security/`:

image


themes/camunda/layouts/shortcodes/ee-download.html
```diff
@@ -45,7 +45,12 @@
           <div class="selected-server">
             <div class="info">
               <h3></h3>
+              <p>
               Released on <span class="date"></span>, <a class="notes">release notes</a>.
+              </p>
+              <p>
+              Camunda Platform 7 Enterprise Edition provides Enterprise components under a license agreement containing restrictions on use and disclosure and that are protected by intellectual property laws (<a class="licenses-link">view licenses</a>). Third-party libraries or application servers included are distributed under their respective licenses (<a class="third-party-licenses-link">view third-party notices</a>).
+              </p>
```
pepopowitz, line 53: Changes in this file came over when I published camunda-docs-theme into this project. Original theme PR: https://github.com/camunda/camunda-docs-theme/pull/27

I'm assuming @ThorbenLindhauer that the intention was to eventually let that theme change propagate to branches here in the docs, so I included it in this PR. Please let me know if I'm incorrect and should remove this change from this PR!

pepopowitz: oh but now I see there is a javascript change also associated with that PR...so I need to figure out how to publish the theme locally so that it minifies the JS/CSS, because I ignored those changes due to a massive diff.

pepopowitz: Figured it out, and my latest commit pulls in the updated minified javascript. 

There were also a handful of CSS changes...but I don't think that was updated as part of the above linked PR, nor do I see any CSS changes in the last couple years....so I ignored those.

akeller: Typically I ignore all changes other than the ones I'm making when working in the C7 docs repos. This does make making PRs more challenging, but I can't always find or figure out who made the changes to ask if the intention is to move them live.

pepopowitz: @akeller I'm not sure if your response is blocking or non-blocking?

pepopowitz:
Heads up that I've force-pushed this branch based on the new approach (also seen in https://github.com/camunda/camunda-docs-manual/pull/1345) for canonicals. It's ready for official review.

themes/camunda/layouts/shortcodes/ee-download.html
akeller: That's going to require a response from the C7 team. I simply don't know if these changes are wanted. Please reach out to Thorben or Tassilo for confirmation.

ThorbenLindhauer: Effectively it doesn't matter for the reader if you include the change in your PR or not, because the short code is only used in camunda-docs-static [here](https://docs.camunda.org/enterprise/download/#full-distributions-and-standalone-web-applications).

I recommend to include it, because the theme must be the source of truth. Changes that go into the theme should eventually roll out to repositories and branches that consume the theme. It's only that sometime we make changes to the theme that we do not need in all places (like in this instance). Then we may avoid making X PRs to all kinds of branches and repositories.

Please don't ask me why this is defined as a short-code in the theme when it's only used in one place...

pepopowitz:
closing this to reopen with the current approach (versioned canonicals)

PR #1336, opened 2022-11-08
feat: canonicals(7.18) -- fix sitemap & apply rel=canonical strategy to 7.18 branch

pepopowitz:
Part of https://github.com/camunda/product-hub/issues/232.

1. Updates the theme based on https://github.com/camunda/camunda-docs-theme/pull/31
2. Fully qualifies the `baseURL` setting so that sitemap is valid, and rel=canonical can be set
3. Applies the rel=canonical strategy discussed in https://github.com/camunda/product-hub/issues/232#issuecomment-1292543102
4. Updates the sitemap to emit true canonicals for this version -- URLs that don't contain a version.

## Proof that the sitemap gets generated with absolute & versionless URLs

I ran a build locally and this is the generated sitemap:

image

## Proof that pages are generated with a versionless rel=canonical

I ran a build locally and this is the generated page for `/manual/7.18/user-guide/security/`:

image


akeller (approved):
This looks good to me. Since this is a theme change, I know you picked up some of the other changes that were in the theme repo but were unpublished. I don't see any (obvious) negative consequences from pushing these changes.

PR #1344, opened 2022-11-11
chore(rest): move to OpenAPI REST documentation

mboskamp:
related to camunda/camunda-bpm-platform#2876


content/reference/rest/specification.md
```diff
@@ -0,0 +1,12 @@
...
+title: 'REST API specification'
+weight: 20
+
+menu:
+  main:
+    name: "Specification"
+    identifier: "rest-api-spec"
+    parent: "rest-api"
+
+---
```
tasso94, line 12:
```suggestion
---

On production, this should redirect to the following page: 

* https://stage.docs.camunda.org/rest/camunda-bpm-platform/{{< minor-version >}}/

Or to this page in case you are viewing the docs of the `master` or `latest` branch:

* https://stage.docs.camunda.org/rest/camunda-bpm-platform/{{< minor-version >}}-SNAPSHOT/

```
Let's give some context in case a developer stumbles upon this page.

tasso94 (changes requested):
Great job! Looks good to me in general. Just some hints added below ⬇️

content/reference/rest/specification.md
mboskamp: Good point, I think we should also add it to the [javadoc](https://github.com/camunda/camunda-docs-manual/blob/master/content/reference/javadoc.md) page.

content/reference/rest/specification.md
```diff
@@ -0,0 +1,12 @@
...
+
+title: 'REST API specification'
+weight: 20
+
+menu:
+  main:
+    name: "Specification"
```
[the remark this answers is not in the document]
mboskamp: Now thinking about it a bit more, doesn't the arrow usually indicate that a link opens in a new tab? Also, I think that would be the better behavior here since there is no real way to get back to the camunda docs from the REST documentation. The same applies to the Javadocs link. Both are currently opening in the same tab.

content/reference/rest/specification.md
mboskamp: Is there a reason why you chose the stage version of the docs for the link?

content/reference/rest/specification.md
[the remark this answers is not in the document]
tasso94: The usual pattern is that an arrow-up icon indicates an external link. Opening it in a new tab is an implementation detail of that pattern. Cf. behavior of Wikipedia for instance: 
image

How can we technically achieve that the page opens in a new tab? With the arrow icon, a user can at least use right-click or a shortcut to open the page in a new tab.

content/reference/rest/specification.md
tasso94: No.

content/reference/rest/specification.md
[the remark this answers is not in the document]
tasso94: I agree. 🙂👍

PR #1382, opened 2023-01-05
docs : update apache http client url

marunrun:
update apache http client url


danielkelemen (approved):
Looks good.
Thank you @marunrun for the fix!

PR #1392, opened 2023-01-19
Tech writer review: docs(multi-tenancy): clarify access to tenants as an admin

tasso94:
related to camunda/camunda-bpm-platform#3079


christinaausley (approved):
Minor adjustment in latest commit to change "he" to "they". LGTM 👍

PR #1395, opened 2023-01-24
experiment: remove noindex tag for a few pages

pepopowitz:
Associated with https://github.com/camunda/developer-experience/issues/22. See that issue for a higher level view of all the steps associated with this experiment.

Configures 3 experimental URLs to exclude the `noindex` tag, so that we can see what effect that has on Google's selection of canonical URLs. 

* Google's sees the canonical for each of these 3 pages as the 7.17 version, instead of 7.18.
* All 3 of these pages already include a `link rel=canonical` tag pointing at the 7.18 version.

Once merged & deployed, I'll submit all three pages to Google to reindex. After a period of time (a month or so?), I'll reassess the Google canonical selection for the pages, to see if anything improved.

If things improve for all 3 pages (e.g. Google instead chooses the 7.18 version as canonical), we can talk about how to introduce to a larger experiment set. 

The worst case scenario, as I see it, is that Google will still choose the 7.17 page as canonical, however that page will now be in the search index, and users will get the 7.17 page in their search results instead of no page.

**Note**: I did not add these changes to the theme, as they are temporary, and I only really care if they're in this version right now. If the experiment proves that we want the changes in other versions, I'll start by applying them to the theme.

## Proof that it emits the `noindex` tag for pages other than the 3 in the experiment group

image

## Proof that it excludes the `noindex` tag for the 3 pages in the experiment group

image

image

image

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -9,6 +9,10 @@ blackfriday:
 
 params:
   canonicalBasePath: "https://docs.camunda.org"
+  skipNoIndexPages:
+    - "/manual/7.17/reference/rest/case-instance/variables/get-variable-binary/"
+    - "/manual/7.17/reference/rest/process-definition/delete-process-definition/"
+    - "/manual/7.17/update/minor/71-to-72/glassfish/"
   sections:
     - id: "get-started"
       name: "Get Started"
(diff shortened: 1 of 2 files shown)


akeller (approved):
> Note: I did not add these changes to the theme, as they are temporary, and I only really care if they're in this version right now. If the experiment proves that we want the changes in other versions, I'll start by applying them to the theme.

I think this makes complete sense for an experiment. 

This looks good and I'm crossing all my fingers and toes this gives some some clues. And thanks again for pushing through this.

PR #1402, opened 2023-02-20
Update input.md

PhyberApex:
JUEL has been replaced with FEEL since 7.13


tasso94:
Hi @PhyberApex,

Can you please create your PR with the base branch `camunda:master` instead of `camunda:7.17`? Otherwise, your change will never be included in newer versions of the Camunda 7 docs.

Best,
Tassilo

PhyberApex:
Did so over at https://github.com/camunda/camunda-docs-manual/pull/1407

PR #1405, opened 2023-02-22
chore(engine): add documentation about set retries with due date

mboskamp:
Related to camunda/camunda-bpm-platform/issues/2751


content/user-guide/process-engine/the-job-executor.md
```diff
@@ -532,7 +532,7 @@ The retry times would be three and the behaviour for this example would be the f
...
 * A job fails for the fourth time: the job will **NOT** be retried again and the next due date is in 20 minutes (PT20M is applied again).
 
-If the user decides to increase the retry number during retries, the last interval of the list would be applied within the difference of the new value and the size of the list. After that it would continue with the normal flow as above.
+If the user decides to increase the retry number during retries, the last interval of the list would be applied within the difference of the new value and the size of the list. After that it would continue with the normal flow as above. The APIs to increase the number of retries also supports to pass a new due date for the job which will determine when the job will be picked up again. This is useful in cases where you want to control when a failed job should be tried again. For example, an incident has been resolved and the job should run as soon as possible instead of waiting for the next retry cycle. In this case setting the job due date to the current date or a date in the past would make the job executor pick up the job immediately.
```
tmetzke, line 535:
```suggestion
If the user decides to increase the retry number during retries, the last interval of the list would be applied within the difference between the new value and the size of the list. After that, it would continue with the normal flow as above. The API to increase the number of retries also supports passing a new due date for the job which will determine when the job will be picked up again. This is useful in cases where you want to control when a failed job should be tried again. For example, an incident has been resolved and the job should run as soon as possible instead of waiting for the next retry cycle. In this case, setting the job due date to the current date or a date in the past would make the job executor pick up the job immediately.
```

tmetzke:
The current changes look good. Let's also incorporate [what was suggested](https://github.com/camunda/camunda-bpm-platform/issues/3167#issuecomment-1440316606) in the related issue.

tmetzke (approved):
Awesome 👍

PR #1407, opened 2023-02-22
Replacing JUEL with FEEL as referenced to be default

PhyberApex:
JUEL has been replaced with FEEL since 7.13


tasso94:
Looks good to me. Thanks!

tasso94 (approved):

PR #1409, opened 2023-02-24
experiment: remove `noindex` tag from 7.15 pages

pepopowitz:
This experiment is explained in detail in https://github.com/camunda/developer-experience/issues/22#issuecomment-1444447197.

Removes the `meta noindex` tag from 7.15 pages, to see if this (a) has a positive effect of correcting any incorrectly canonicalized 7.15 pages, and (b) does not have the negative effect of prompting Google to index 7.15 pages instead of 7.18 pages. 

Note that these pages _do_ have a canonical URL pointing at the 7.18 version, which I expect to protect us from Google canonicalizing _and_ indexing these 7.15 pages.

I plan to revisit this before the next product release, and straighten things out at the theme level instead of only in this specific version.

## Proof that it removes the `noindex` tag

image


akeller (approved):
> Note that these pages do have a canonical URL pointing at the 7.18 version, which I expect to protect us from Google canonicalizing and indexing these 7.15 pages.

Very important call out. No concerns from me.

PR #1414, opened 2023-03-02
chore(user-guide): bump quarkus versions

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/2788


tmetzke (approved):
👍 Looks good to me in general.
🔧 Let's move to version declarations without a specific patch level.

PR #1431, opened 2023-03-28
fix(element-templates): redirect to correct docs

ManuelDittmar:
- previous url results in a 404


tasso94 (approved):
Thanks for your contribution! 🚀

PR #1434, opened 2023-03-30
fix(introduction): fix license book links

ThorbenLindhauer:
related to camunda/camunda-bpm-platform#3296

### DIFF
--- a/content/introduction/third-party-libraries/camunda-bpm-platform-license-book.md
+++ b/content/introduction/third-party-libraries/camunda-bpm-platform-license-book.md
@@ -2316,7 +2316,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 
 </pre></details></div>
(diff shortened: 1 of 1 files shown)


tmetzke:
🙃 Now that you brought this to my attention, I can see some other weird links we might want to correct, like `git:`, `git+ssh:`, and `git+https:`.

tmetzke (approved):
This should be the only required ones 👍 
I added a comment above for optional other links to adjust.

ThorbenLindhauer:
I'd like to only do the minimal thing to make the build succeed (e.g. any manual fix will be overridden with the next minor release again). If we wanted to improve the other links, we should rather look at how we generate the Markdown. But I don't think that's worth our time at the moment.

tmetzke:
> I'd like to only do the minimal thing to make the build succeed (e.g. any manual fix will be overridden with the next minor release again). If we wanted to improve the other links, we should rather look at how we generate the Markdown. But I don't think that's worth our time at the moment.

Makes sense 👍

PR #1446, opened 2023-04-17
improve(user-guide): clarify end event of execution listeners

ThorbenLindhauer:
related to camunda/camunda-bpm-platform#3046


tasso94 (approved):
👍

PR #1451, opened 2023-04-22
chore(update): add 7.19 to 7.20 guide with OLE on Postgres

tmetzke:
related to https://github.com/camunda/camunda-bpm-platform/issues/3214


tmetzke:
@christinaausley, please note that the content of the following files is not really new but rather a copy of the previous update guides. Nonetheless, it might be good to involve you here now so we can improve those going forward.

* `content/update/minor/719-to-720/jboss.md`
* `content/update/minor/719-to-720/tomcat.md`
* `content/update/minor/719-to-720/was.md`
* `content/update/minor/719-to-720/wls.md`

The same goes for everything except the last section in `content/update/minor/719-to-720/_index.md`.

Depending on the number of changes you would recommend for those files, we can also think about backporting those changes to the update guides of all the currently supported versions. But this is optional.

yanavasileva (approved):
👍 Looks good to me.

christinaausley (approved):
Nice work here 👍

PR #1452, opened 2023-04-22
chore(update): add OLE on Postgres to patch level guide

tmetzke:
related to https://github.com/camunda/camunda-bpm-platform/issues/3214


yanavasileva (approved):
👍 Looks good to me.

PR #1456, opened 2023-05-03
chore(update): add hint for JUEL method resolution

tmetzke:
related to https://github.com/camunda/camunda-bpm-platform/issues/3053

### DIFF
--- a/content/update/patch-level.md
+++ b/content/update/patch-level.md
@@ -414,7 +414,7 @@ Previoulsy, overloaded methods couldn't be determined reliably and were chosen b
 
 Now, the most specific version according to the provided number and types of method parameters is chosen consistently. For example, method `myMethod` expecting an `Integer` is chosen over method `myMethod` expecting an `Object` if the provided parameter is an `Integer` or can be coerced into one.
 
-Ideally, you shouldn't notice any difference in method invocation from JUEL expressions. However, we recommend testing your existing expressions thoroughly before using version in production.
+Ideally, you shouldn't notice any difference in method invocation from JUEL expressions. However, we recommend testing your existing expressions thoroughly before using this patch version in production.
 
 # Full Distribution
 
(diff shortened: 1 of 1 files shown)


tmetzke:
I am not entirely sure if this creates more uncertainty than necessary. Looking for feedback. 🙂

content/update/patch-level.md
```diff
@@ -404,6 +404,18 @@ With this release, we adjusted the [Optimistic Locking]({{< ref "/user-guide/pro
...
+
+### Bean method resolution from JUEL expressions
+
+Calling bean methods from JUEL expressions now works more reliably and supports overloaded methods. As a result, the behavior of calling overloaded methods from such expressions changes.
+
+Previoulsy, overloaded methods couldn't be determined reliably and were chosen based on the order of the underlying JDK's result of the call to `Class#getMethods` for a specific method name. As a result, the expected method might have been chosen for the expression evaluation but this could not be guaranteed.
+
+Now, the most specific version according to the provided number and types of method parameters is chosen consistently. For example, method `myMethod` expecting an `Integer` is chosen over method `myMethod` expecting an `Object` if the provided parameter is an `Integer` or can be coerced into one.
+
+Ideally, you shouldn't notice any difference in method invocation from JUEL expressions. However, we recommend testing your existing expressions thoroughly before using version in production.
```
tasso94, line 417:
```suggestion
Ideally, you shouldn't notice any difference in method invocation from JUEL expressions. However, we recommend testing your existing expressions thoroughly before using this patch version in production.
```

tasso94 (approved):
> I am not entirely sure if this creates more uncertainty than necessary.

I think making this transparent creates value:

* If the changed behavior breaks the tests of a user unexpectedly, they can easily understand that this might be because we changed the behavior slightly. 
* In the past I saw JDKs that behave differently then others. Maybe there exist JDKs that reliably returned the wrong method and users took that for granted. I don't expect this happens for the majority of users. However, those that struggle can learn on their own how to tweak their implementation instead of creating support cases which ideally means less support load.

tmetzke:
> > I am not entirely sure if this creates more uncertainty than necessary.
> 
> I think making this transparent creates value:
> 
> * If the changed behavior breaks the tests of a user unexpectedly, they can easily understand that this might be because we changed the behavior slightly.
> * In the past I saw JDKs that behave differently then others. Maybe there exist JDKs that reliably returned the wrong method and users took that for granted. I don't expect this happens for the majority of users. However, those that struggle can learn on their own how to tweak their implementation instead of creating support cases which ideally means less support load.

Thanks, @tasso94, this is really helpful 👍

christinaausley (approved):
Nice 👍 No adjustments from me on this one.

yanavasileva (approved):

PR #1476, opened 2023-06-09
docs(migration): document container-based auth requirements

tasso94:
related to camunda/camunda-bpm-platform#3500


yanavasileva (approved):
👍

PR #1488, opened 2023-07-03
docs(webapps): add documentation about login plugin points

tasso94:
related to camunda/camunda-bpm-platform#3412


danielkelemen (approved):
👍 Sounds good.

PR #1503, opened 2023-07-26
chore(rpa): remove RPA support for 7.20

mboskamp:
Related to https://github.com/camunda/camunda-bpm-platform/issues/3606


psavidis:
❓ Do we need to keep the [RPA License Book](https://docs.camunda.org/manual/latest/introduction/third-party-libraries/camunda-bpm-rpa-bridge-license-book/) & its reference under _licenses_ for version **7.20**?

mboskamp:
@psavidis, you are right, I think the license book for the RPA Bridge can go. I removed it only on master.

psavidis (approved):
I think there is nothing else left for 7.20 and all other references of RPA Bridge are referred only in versions <= 7.19.

Good job 👍

PR #1504, opened 2023-07-27
feat: update link to c7 element templates docs

barmac:
Related to https://github.com/camunda/camunda-platform-docs/pull/2369

### DIFF
--- a/content/modeler/element-templates.md
+++ b/content/modeler/element-templates.md
@@ -12,4 +12,4 @@ menu:
 
 ---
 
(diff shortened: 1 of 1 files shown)


barmac:
This will be a draft until the linked PR is approved due to tight coupling with links.

tmetzke:
Thanks, @barmac, just ping me when the upstream change is done. I'll check the link then and approve the change so we can merge it right after 👍

barmac:
Since we decided against having a dedicated page for C7 element templates, and only "Defining templates in C7" was extracted, there is no need for a change. I am closing this PR.

PR #1505, opened 2023-07-27
chore(docs): fix broken webapps links

danielkelemen:
related to camunda/camunda-bpm-platform#3393

### DIFF
--- a/content/user-guide/process-engine/process-diagram-api.md
+++ b/content/user-guide/process-engine/process-diagram-api.md
@@ -40,4 +40,4 @@ viewer.importXML(xml, function(err) {
 });
 ```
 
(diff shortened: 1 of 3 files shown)


psavidis:
I've tested the new link and it works. LGTM 👍

psavidis (approved):

psavidis:
❓ Does this change need to be backports to other versions e.g **7.19**, **7.18**, **7.17**?

danielkelemen:
> ❓ Does this change need to be backported to other versions e.g **7.19**, **7.18**, **7.17**?

It could be in 7.19 but I wouldn't bother. These are just super minor things and they have been there for a while now.

psavidis (approved):
If the effort is low, personally i think it makes sense to backport the change of the docs to the previous supported versions. It's documentation pointing nowhere. Feel free to proceed as you like with this decision.

PR #1506, opened 2023-07-28
Provide a better comment on Identity Provider

jonathanlukas:
The interface `ReadOnlyIdentityProvider` does not contain the methods `#findUserByQueryCriteria` and `#findGroupByQueryCriteria`. 

The actual requirement for the `AuthenticationUtil` class to function is that 
* the user can be found
* the groups can be queried
* the tenants can be queried

By default, this is already the case. The only known cases where we are facing problems is when the container-based authentication is used to perform SSO.


tasso94:
Hi @jonathanlukas,

Thanks for pointing this out. You are right!
I adjusted your wording slightly and merged it.

Best,
Tassilo

PR #1515, opened 2023-08-08
[Acquisition team] Add buttons to navbar

akeller:
Follow up to https://github.com/camunda/camunda-docs-theme/pull/41 and adding the changes to this part of the docs site.

When running locally, two additional shortcode files were added, but they do not appear to be part of this work and are not in this PR.


themes/camunda/assets/css/docs.css
pepopowitz, line 1: The large diffs on these assets (css and js) make me think they were not minified. Did you run the `optimize` command [described in the theme docs](https://github.com/camunda/camunda-docs-theme#consuming-the-theme)?

> 2) Build and update the theme in the target location by either running:
>    1) `grunt build` for building and syncing non-minified assets
>    2) Run additionally `grunt optimize` for building and syncing minified assets

pepopowitz (changes requested):
I think the assets should be minified, otherwise this looks great!

themes/camunda/assets/css/docs.css
akeller: Fixed, thank you!

tasso94 (approved):
Looks good to me! 👍

PR #1517, opened 2023-08-15
chore(introduction): add wildfly 29 to supported environments

mboskamp:
Related to https://github.com/camunda/camunda-bpm-platform/issues/3587


tmetzke (approved):
Good catch with version 23 👍

PR #1518, opened 2023-08-18
docs(bpmn): re-evaluate a time cycle

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/2570

### DIFF
--- a/content/reference/bpmn20/events/timer-events.md
+++ b/content/reference/bpmn20/events/timer-events.md
@@ -105,6 +105,25 @@ You can use expressions for the timer event definitions. By doing so you can inf
 </boundaryEvent>
 ```
 
+### Re-evaluate a time cycle
+
+A timer's repeat cycle can be updated if the defined expression is changed. This is achieved by re-evaluation of the expression next time when a timer is fired. So the new cycle will be affective with next scheduled timer.
+
+Let's observe a scenario with the following timer defined where `myBean.getCycle()="R3/PT2H"`:
+
+```xml
+<startEvent id="theStart">
+  <timerEventDefinition>
+    <timeCycle>#{myBean.getCycle()}</timeCycle>
+  </timerEventDefinition>
+</startEvent>
+```
+Timer is meant to be fired at 1 p.m., 3 p.m., and 5 p.m.
+
+At 4 p.m., we adjust to return `myBean.getCycle()="R2/PT30M"`. Then timer will be triggered at 5 p.m. (still calculated by the previous cycle), 5:30 p.m. and 6 p.m. (based on the new cycle).
+
+The feature is disabled by default, to enable the feature set the `reevaluateTimeCycleWhenDue` property to `true` in the process engine configuration.
+
 ## Handling of Timezones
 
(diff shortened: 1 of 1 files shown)


content/reference/bpmn20/events/timer-events.md
```diff
@@ -105,6 +105,25 @@ You can use expressions for the timer event definitions. By doing so you can inf
 </boundaryEvent>
 ```
 
+### Re-evaluate a time cycle
+
+A timer's repeat cycle can be updated if the defined expression is changed. This is achieved by re-evaluation of the expression next time when a timer is fired. So the new cycle will be affective with the next scheduled timer.
```
tmetzke, line 110:
```suggestion
A timer's repeat cycle can be updated if the defined expression is changed. This is achieved by re-evaluation of the expression next time when a timer is fired. So the new cycle will be effective from the next scheduled timer.
```

content/reference/bpmn20/events/timer-events.md
```diff
@@ -105,6 +105,25 @@ You can use expressions for the timer event definitions. By doing so you can inf
...
+Let's observe a scenario with the following timer defined where `myBean.getCycle()="R3/PT2H"`:
+
+```xml
+<startEvent id="theStart">
+  <timerEventDefinition>
+    <timeCycle>#{myBean.getCycle()}</timeCycle>
+  </timerEventDefinition>
+</startEvent>
+```
+Let's consider the timer is meant to be fired at 1 p.m., 3 p.m., and 5 p.m.
```
tmetzke, line 121:
```suggestion
Let's assume the timer is meant to fire at 1 p.m., 3 p.m., and 5 p.m.
```

content/reference/bpmn20/events/timer-events.md
```diff
@@ -105,6 +105,25 @@ You can use expressions for the timer event definitions. By doing so you can inf
...
+
+At 4 p.m. after the timer has been fired two times already, we adjust the bean to return `myBean.getCycle()="R2/PT30M"`. Then timer will still be triggered at 5 p.m. (as initially scheduled calculated by the previous cycle), and afterward at 5:30 p.m. and 6 p.m. (based on the new cycle).
```
tmetzke, line 123:
```suggestion
At 4 p.m., after the timer has fired two times, we adjust the bean to return `myBean.getCycle()="R2/PT30M"`. In effect, the timer will still be triggered at 5 p.m. (as initially scheduled and calculated by the previous cycle). Afterward, it will fire at 5:30 p.m. and 6 p.m. (based on the new cycle).
```

content/reference/bpmn20/events/timer-events.md
```diff
@@ -105,6 +105,25 @@ You can use expressions for the timer event definitions. By doing so you can inf
...
+
+The feature is disabled by default, to enable it, set the `reevaluateTimeCycleWhenDue` property to `true` in the process engine configuration.
```
tmetzke, line 125:
```suggestion
The feature is disabled by default. To enable it, set the `reevaluateTimeCycleWhenDue` property to `true` in the process engine configuration.
```

tmetzke (approved):
👍 Very good description, I like the use of the example.
🔧 I added some minor grammar suggestions.

PR #1520, opened 2023-08-22
docs(sec): add paragraph on how to prevent bypassing authorizations

tasso94:
related to https://github.com/camunda/camunda-bpm-platform/pull/3692


danielkelemen (commented):
💬 I think _writable_ is also correct.
❓ You also changed the javadoc reference but didn't rename the class. It's still [WritableIdentityProvider](https://docs.camunda.org/javadoc/camunda-bpm-platform/7.19/org/camunda/bpm/engine/impl/identity/WritableIdentityProvider.html). Or am I missing something?
👍 The newly added paragraph looks good though!

tasso94:
Re Writable vs Writable. I thought this was just a typo in the docs, but you are right. We use Writable in our source code. I'll revert it back to Writable.

danielkelemen (approved):
👍

PR #1522, opened 2023-08-29
Delete FEEL version number

xomiamoore:
Per Slack, this is outdated and difficult to keep updated, so I deleted the version number.


tmetzke (commented):
Thanks, @xomiamoore, for creating a PR for this 👍 
If I'm not mistaken, there also is a version reference in line 27. I guess we should remove that as well, what do you think?

xomiamoore:
@tmetzke That refers to FEEL Scala, I'm not sure if the statement `By default, Camunda Platform relies on the FEEL Scala Engine in version 1.11.0.` is still correct in current versions. But if that updates frequently too, we should remove it for sure.

tmetzke:
@xomiamoore, let me quickly check how/if we updated this in the past. I think we should remove the version reference if we did.
I also added some reasoning for this in the tracking issue: https://github.com/camunda/camunda-bpm-platform/issues/3723

tmetzke:
Looking at it more closely, it's actually the FEEL Scala engine version that we should remove.
Since I cannot find much documentation about the FEEL version itself (which I think refers to the DMN version level of FEEL) in our docs, we could think about removing it to avoid confusion.

@ThorbenLindhauer, would you agree with that assessment or do you think we should keep the FEEL 1.2 version reference for any reason?

ThorbenLindhauer:
@tmetzke I suggest to remove the FEEL version, too. The main use of the mentioned versions is that users can understand what FEEL features they can use in their models. The FEEL version doesn't help with that, because feel-scala doesn't indicate that. The feel-scala version would help, but we have no good way of keeping it up to date, so I'd remove it to because otherwise it leads to more confusion than benefit. Assuming that the C7 feel-scala version is never too much behind the latest, I'd think that the practical problem of not knowing the exact version is not very big.

tmetzke:
Totally agree, @ThorbenLindhauer, I'll take care of it and adjust the docs 👍

tmetzke:
@xomiamoore I removed the FEEL Scala Engine version as well now. I gave the other sections a second look, their appear still valid to me. I'll go ahead and merge the changes. I'll quickly also backport them since that helps and is easy to do.

PR #1534, opened 2023-09-20
chore(wildfly): add wildfly 29 section to migration guide

mboskamp:
Related to https://github.com/camunda/camunda-bpm-platform/issues/3587


mboskamp:
You are probably right. I added it because the issue description said so. I did not ask myself if it was really necessary. Let's not merge this PR.

PR #1542, opened 2023-09-28
chore(database): update erd

danielkelemen:
related to camunda/camunda-bpm-platform#3336


tmetzke (changes requested):
👍 Nice catch, @danielkelemen.
❌ Let's create `7.19` and `7.20` files here (can be exact copies). Looks like this was missed with 7.19 and should also be backported correctly. Backporting can happen in a follow-up ticket. Backporting can also include the change you made here.

Background:
Every version [has its own file](https://confluence.camunda.com/pages/viewpage.action?spaceKey=AP&title=Performing+a+Minor+Release#PerformingaMinorRelease-UpdatetheEntity-RelationshipDiagrams), no matter if the content is the same. That way, patches can be incorporated easily later on, which is quite a hassle otherwise.

danielkelemen:
> Background: Every version [has its own file](https://confluence.camunda.com/pages/viewpage.action?spaceKey=AP&title=Performing+a+Minor+Release#PerformingaMinorRelease-UpdatetheEntity-RelationshipDiagrams), no matter if the content is the same. That way, patches can be incorporated easily later on, which is quite a hassle otherwise.

In that case I'll create 7.19 and 7.20 copies from all the `.mwb` and SVG files.

tmetzke (approved):
Awesome 👍

PR #1554, opened 2023-10-22
Update database-schema.md

Ironarrow78:
For the migration to Liquibase it is important to run the changelogSyncToTag command from a special folder location. Otherwise Liquibase doesn't recognize the processed SQL files.

(--> https://forum.camunda.io/t/problems-migrating-to-liquibase/45798)

### DIFF
--- a/content/installation/database-schema.md
+++ b/content/installation/database-schema.md
@@ -149,6 +149,10 @@ Find the row with the highest value in the `ID_` column and use the value of thi
 1. Run Liquibase's [changelogSyncToTag command](https://docs.liquibase.com/commands/community/changelogsynctotag.html) referencing the `camunda-changelog.xml` and using your current database schema version as the tag.
 You can pass on the connection details to your database instance via parameters as described in the Liquibase documentation or [create a properties file](https://docs.liquibase.com/workflows/liquibase-community/creating-config-properties.html).
 
+    _Important_: You have to run the changelogSyncToTag command from the `{basedir}\src\main\resources` folder. Otherwise Liquibase doesn't recognize the processed SQL files.
+
+
+
 Liquibase uses this information to create the tracking tables and mark all changesets until the tag you defined as executed.
 Liquibase determines if there are any changes that it needs to apply to your database for any subsequent `update` commands.
 You have migrated your manual installation to Liquibase.


mboskamp (changes requested):
Hi @Ironarrow78,
Thank you for pointing out this flaw in the documentation. Thanks again for going the extra mile and opening this PR.

When reading the numbered list of actions that are required for the migration, I think it would make sense to include the hint in the list before the last step. This way, when someone follows the instructions, they get the hint at the right time and not afterward.

My proposal would look something like this:

```
1. Setup Liquibase, e.g. by [downloading Liquibase CLI](https://www.liquibase.org/download)
1. Identify your current database schema version. You can extract this information from the `ACT_GE_SCHEMA_LOG` table.
Find the row with the highest value in the `ID_` column and use the value of this row's `VERSION_` column.
1. Navigate to `{basedir}\src\main\resources` folder.
1. Run Liquibase's [changelogSyncToTag command](https://docs.liquibase.com/commands/community/changelogsynctotag.html) referencing the `camunda-changelog.xml` and using your current database schema version as the tag.
You can pass on the connection details to your database instance via parameters as described in the Liquibase documentation or [create a properties file](https://docs.liquibase.com/workflows/liquibase-community/creating-config-properties.html).
```

What do you think?
Another point that we should address is porting this change to at least the currently supported versions. Right now, the change would only be visible for 7.20 and not for 7.21(master) or versions before 7.20.

mboskamp:
Hi @Ironarrow78, we have not heard back from you for some time. I will close this ticket for now. If you want to drive the issue further, please don't hesitate to open a new ticket or reply here.

Cheers,
Miklas

Ironarrow78:
@mboskamp Sorry for the late reply. I've had some trouble with my mail account so I couldn't read your messages.
For me it's fine to change the documentation like you written above.

Best regards,
Rainer

mboskamp:
Hi @Ironarrow78,
Would you be up to creating pull requests for the relevant branches? Currently maintained versions are `master`, `7.20`, `7.19`, and `7.18`.

Cheers,
Miklas

Ironarrow78:
Hi @mboskamp ,
I made the changes to 7.18, 7.19 and the master. I hope this is all you need.

Best regards,
Rainer

mboskamp:
Hi @Ironarrow78,
I can't see any other PRs from you in this project. Could you double-check? Also, this PR does not have the suggested improvements yet.

mboskamp:
@Ironarrow78, I haven't received an update in a while now. I will close the PR and related issue for you.

PR #1558, opened 2023-10-27
chore/fix: update theme to prevent search results from masking social buttons (master)

pepopowitz:
Follow-up to https://github.com/camunda/camunda-docs-theme/pull/47. 

Incorporates the theme changes from that PR into the `master` branch. This will prevent the undisplayed search-results window from masking the social buttons at the top of the screen.


yanavasileva (approved):
👍 Looks good, tried it out locally.

PR #1562, opened 2023-11-08
docs(process-engine): Add Documentation to Fetch External Task

psavidis:
- Rename Fetch External Task Section
- Split it into individual sections for describing sorting by priority, createTime & multi-level sorting

Related-to: https://github.com/camunda/camunda-bpm-platform/issues/3929, https://github.com/camunda/camunda-bpm-platform/issues/3896

### DIFF
--- a/content/user-guide/process-engine/external-tasks.md
+++ b/content/user-guide/process-engine/external-tasks.md
@@ -402,6 +402,13 @@ ExternalTask1 [priority=0, createTime=1]
 
 Note: The `createTime` field used in the example uses numbers for easing the visual demonstration. In real results, the `createTime` will be populated using a `Date` value.
 
+{{< note title="" class="info" >}}
+It is likely more frequent that multiple tasks might share the same priority compared to their createTime. As a result, priority makes sense to be used
+as a first sorting criterion and then sort by createTime on priority equality. Therefore, priority will always have precedence over createTime during
+fetch and lock sorting configurations.
+{{< /note >}}
+
+
 ### Completing Tasks
 
 After fetching and performing the requested work, a worker can complete an external task by calling the `ExternalTaskService#complete` method. A worker can only complete tasks that it fetched and locked before. If the task has been locked by a different worker in the meantime, an exception is raised.
--- a/content/user-guide/process-engine/external-tasks.md
+++ b/content/user-guide/process-engine/external-tasks.md
@@ -403,9 +403,7 @@ ExternalTask1 [priority=0, createTime=1]
 Note: The `createTime` field used in the example uses numbers for easing the visual demonstration. In real results, the `createTime` will be populated using a `Date` value.
 
 {{< note title="" class="info" >}}
-It is likely more frequent that multiple tasks might share the same priority compared to their createTime. As a result, priority makes sense to be used
-as a first sorting criterion and then sort by createTime on priority equality. Therefore, priority will always have precedence over createTime during
-fetch and lock sorting configurations.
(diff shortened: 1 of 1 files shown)


content/user-guide/process-engine/external-tasks.md
```diff
@@ -341,7 +342,64 @@ for (LockedExternalTask task : tasks) {
...
+  String topic = task.getTopicName();
+
+  // work on task for that topic
+  ...
+}
+```
+
+You can update external task rows in the database with a custom create time to influence the ordering of external tasks created with versions < 7.21.0.
+
+#### Multi-Level Sorting
```
mboskamp, line 377: 🔧 Should we add a paragraph explaining that `priority` always takes precedence before all other sorting options?

mboskamp (changes requested):
👍 I did a review of the technical content. Feel free to assign a technical writer for review as well, if you like.

content/user-guide/process-engine/external-tasks.md
psavidis: I've added a separate note paragraph to explain the priority precedence and its reasoning behind it.

psavidis:
Hi @christinaausley,

This commit complements the documentation with the new sorting by `createTime` and `multi-level sorting` (combining priority and createTime as sorting options) to the external tasks section.

Would you like to have a look and suggest possible changes or approve if everything looks good to you?
The section that is changed is : `Fetch External Task` | **By Priority** | **By Create Time**.

Your feedback would be much appreciated :)

content/user-guide/process-engine/external-tasks.md
```diff
@@ -341,6 +342,71 @@ for (LockedExternalTask task : tasks) {
...
+ExternalTask3 [priority=0, createTime=3]
+ExternalTask1 [priority=0, createTime=1]
+```
+
+Note: The `createTime` field used in the example uses numbers for easing the visual demonstration. In real results, the `createTime` will be populated using a `Date` value.
+
+{{< note title="" class="info" >}}
+It is likely more frequent that multiple tasks might share the same priority compared to their createTime. As a result, priority makes sense to be used
+as a first sorting criterion and then sort by createTime on priority equality. Therefore, priority will always have precedence over createTime during
+fetch and lock sorting configurations.
```
mboskamp, line 408: 🔧 This sounds like the reason priority is a stronger sorting field over other fields is probability. I don't think that is accurate. The reason is that the priority feature is meant to be used to sort the external tasks by the importance of immediate execution. But I don't think this is something we need to explain here. I would simply mention that it takes precedence over any other sorting property.
🔧 This section is not about `createTime` but multi-level sorting. We should keep it more general, so it does not become inaccurate if we ever add more sorting options.

psavidis: Just added a sentence that priority has precedence over createTime.

If we ever add another field, we can add one more sentence in the pattern "createTime has precedence over {newField}" and so on and so forth.

mboskamp: I don't like that. In the implementation, we made efforts to make our future lives easier by choosing an extendable approach. Why shouldn't we just say, "Priority will always take precedence over any other sorting property."? It will always be true in the future. I think the multi-level sorting section does not need anything to do with the `createTime` property.

psavidis: I've used your quoted phrase as is.

christinaausley (approved):

mboskamp (approved):

PR #1564, opened 2023-11-10
docs(process-engine): Add skipIoMappings Documentation

psavidis:
Related-to: https://github.com/camunda/camunda-bpm-platform/issues/3934

### DIFF
--- a/content/user-guide/process-engine/batch-operations.md
+++ b/content/user-guide/process-engine/batch-operations.md
@@ -54,11 +54,9 @@ runtimeService.deleteProcessInstancesAsync(processInstanceIds,null, REASON);
 
 There are overloaded methods of the above which allow you to control the following parameters:
 
-`skipCustomListeners` - Skip execution listener invocation for activities that are started or ended as part of this request.
-
-`skipSubprocesses` - Skip deletion of the subprocesses related to deleted processes as part of this request.
-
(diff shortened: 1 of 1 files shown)


content/user-guide/process-engine/batch-operations.md
```diff
@@ -49,10 +49,17 @@ Cancellation of running process instances can be performed asynchronously using
...
+runtimeService.deleteProcessInstancesAsync(processInstanceIds,null, REASON);
 ```
 
+There are overloaded methods of the above which allow you to control the following parameters:
+
+`skipCustomListeners` - Skip execution listener invocation for activities that are started or ended as part of this request.
+
+`skipSubprocesses` - Skip deletion of the subprocesses related to deleted processes as part of this request.
+
+`skipIoMappings` - Skip the IO Mappings if the process instance contains any which would prevent the deletion otherwise.
```
danielkelemen, line 61: 👍 I like the new details about the flags!
🔧 Maybe a list looks better though?

```suggestion
- `skipCustomListeners`: Skip execution listener invocation for activities that are started or ended as part of this request.
- `skipSubprocesses`: Skip deletion of the subprocesses related to deleted processes as part of this request.
- `skipIoMappings`: Skip the IO Mappings if the process instance contains any which would prevent the deletion otherwise.
```

danielkelemen (approved):

PR #1569, opened 2023-11-23
docs(jakarta-binding): added info for camunda-spin-dataformat-xml-dom-jakarta

amardeep2006:


danielkelemen (approved):
Changes look good 👍

danielkelemen:
Hi @amardeep2006,

It makes sense to mention the additional Jakarta library in the docs as latest spring boot and some other distros need to use that.

Thank you for your contribution!

PR #1574, opened 2023-12-20
feat(engine): Introduce operation log for sync message correlations

mboskamp:
Related to https://github.com/camunda/camunda-bpm-platform/issues/3900


danielkelemen (approved):
👍 Nicely written. I just noted two minor things.

PR #1585, opened 2024-01-23
chore(update): add process definition like search change

danielkelemen:
related to camunda/camunda-bpm-platform#4036


tasso94 (approved):
Nice 👍

PR #1607, opened 2024-02-23
chore(update): adjust enterprise link

yanavasileva:
Created on behalf of https://camunda.slack.com/archives/C6M6X3EKZ/p1708616150524839
 the link doesn't occur anywhere else in docs-manual/docs-static

### DIFF
--- a/content/update/ce-to-ee.md
+++ b/content/update/ce-to-ee.md
@@ -12,7 +12,7 @@ menu:
  
 ---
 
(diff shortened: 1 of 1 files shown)


mboskamp (approved):
👍 Looks good. I could not find any more occurrences of the link.
💭 Should we backport this change to 7.20 and 7.19? The redirect works, but I guess we also want the old link to be gone there.

PR #1617, opened 2024-03-13
chore(update): XTaC Java paragraph for jaxb-impl/Jakarta dependencies

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/3829

### DIFF
--- a/content/update/minor/720-to-721/_index.md
+++ b/content/update/minor/720-to-721/_index.md
@@ -79,15 +79,7 @@ This means an update to this version requires code and configuration adjustments
 
 ## Update the client's dependencies
 
-The External Task Client Java has been migrated to the Jakarta namespace. In case you use the default data format providers (e.g.: `DomXmlDataFormatProvider`), you need to bump the version of `com.sun.xml.bind:jaxb-impl` library to 4.0.4:
-
-```xml
-    <dependency>
-      <groupId>com.sun.xml.bind</groupId>
-      <artifactId>jaxb-impl</artifactId>
-      <version>4.0.4</version>
-    </dependency>
-```
+The External Task Client Java has been migrated to the Jakarta namespace. In case you use the default data format providers (e.g.: `DomXmlDataFormatProvider`), you need to update the version of `com.sun.xml.bind:jaxb-impl` dependency to a version >= `4.0.x`.
 
 Complete example of the External Task Client Java can be found [here](https://github.com/camunda/camunda-bpm-examples/tree/7.21/clients/java/order-handling).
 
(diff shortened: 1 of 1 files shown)


content/update/minor/720-to-721/_index.md
```diff
@@ -75,6 +77,20 @@ This means an update to this version requires code and configuration adjustments
...
+
+```xml
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <version>4.0.4</version>
+    </dependency>
+```
+
+Complete example of the External Task Client Java can be found [here](https://github.com/camunda/camunda-bpm-examples/tree/7.21/clients/java/order-handling).
```
psavidis, line 92: ❓ I understand this link will point to the `7.21` tag code once it is released.

Should we use the [link](https://github.com/camunda/camunda-bpm-examples/tree/master/clients/java/order-handling) pointing to `master` ? Otherwise this link will have to be updated for each minor release or am i missing something?

content/update/minor/720-to-721/_index.md
```diff
@@ -75,6 +77,20 @@ This means an update to this version requires code and configuration adjustments
 [HttpClient53]: https://hc.apache.org/httpcomponents-client-5.3.x/index.html
 [HttpClient53-migration]: https://hc.apache.org/httpcomponents-client-5.3.x/migration-guide/index.html
 
+## Update the client's dependencies
+
+The External Task Client Java has been migrated to the Jakarta namespace. In case you use the default data format providers (e.g.: `DomXmlDataFormatProvider`), you need to bump the version of `com.sun.xml.bind:jaxb-impl` library to 4.0.4:
```
psavidis, line 82: ❓ Is it future-proof to state here the exact patch version of the `jaxb` library?

If the dependency is updated in the future (e.g even now in the code review), this reference will have to change.
Would it make sense to have a generic reference to the `jaxb` component that needs to be updated which holds true even if we update the library?
```suggestion
The External Task Client Java has been migrated to the Jakarta namespace. In case you use the default data format providers (e.g.: `DomXmlDataFormatProvider`), you need to bump the `com.sun.xml.bind:jaxb-impl` dependency to a version >= 4.0.x```
```

If the above makes sense, the `xml` snippet can be omitted to avoid the direct reference.

content/update/minor/720-to-721/_index.md
yanavasileva: This is the migration guide to 7.21, so I think it's fine to point to the example with version 7.21.
No need to update it with each release as the change will be relevant only to 7.21.

content/update/minor/720-to-721/_index.md
yanavasileva: Yes, I agree, a generic reference will be better. Will adjust that.

psavidis (approved):

PR #1618, opened 2024-03-14
style(formatting): rename platform 7.20

christinaausley:
Relates to https://github.com/camunda/developer-experience/issues/117.

This is a big chonky PR, but I decided to grind it out in one sitting. Let me know if this review absolutely doesn't work for you and I'll break this up in the subsequent version changes!


akeller (approved):
Thanks for the changes! I saw some list updates as well, which I think are great adjustments. No concerns from me.

PR #1623, opened 2024-03-22
docs(introduction): Update Wildfly Supported Versions

psavidis:
**Context**: Remove 27.0, 29.0 from wildfly supported versions
**Related-to**: https://github.com/camunda/camunda-bpm-platform/issues/3746

### DIFF
--- a/content/introduction/supported-environments.md
+++ b/content/introduction/supported-environments.md
@@ -34,7 +34,7 @@ Run Camunda in every Java-runnable environment. Camunda is supported with our QA
 
 * Apache Tomcat 9.0
 * JBoss EAP 7.4
-* WildFly Application Server 26.0 / 31.0
+* WildFly Application Server 23.0 / 26.0 / 31.0
 * IBM WebSphere Liberty 22 with Java EE 8 features ([Enterprise Edition only](http://camunda.com/enterprise/), 
   see [Installation guide]({{< ref "/installation/full/was/manual-liberty.md" >}}))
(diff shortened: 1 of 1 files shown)


content/introduction/supported-environments.md
```diff
@@ -34,7 +34,7 @@ Run Camunda in every Java-runnable environment. Camunda is supported with our QA
 
 * Apache Tomcat 9.0
 * JBoss EAP 7.4
-* WildFly Application Server 23.0 / 26.0 / 31.0
```
psavidis: ❓ Do we support wildfly 23 anymore? I couldn't find any reference in the code; only the latest (31) and 27 compatibility versions.

yanavasileva: We still support WildFly 23 (used for JBoss EAP 4 that we still support too). We don't have test for it as the code for WildFly 26 can be used to spin Camunda.
Dropping support WildFly 23 will be considered in future release.

content/introduction/supported-environments.md
yanavasileva, line 1: ❓ Shouldn't we remove the CRDB from this page too? Otherwise users might be confused.
`CockroachDB v20.1.3 ` (line 59)

❓ The ticket mentions "only remove CI; don't remove code yet". But I think we can already clean up the docs too during the code freeze. (Search for `Cockroach` under `/content/*.md`)

psavidis: Sure. Removed all cockroachdb references in a separate [PR](https://github.com/camunda/camunda-docs-manual/pull/1625)

yanavasileva (approved):

PR #1632, opened 2024-04-05
docs(xtac): Add useCreateTime and orderByCreateTime properties

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4109

### DIFF
--- a/content/user-guide/ext-client/spring-boot-starter.md
+++ b/content/user-guide/ext-client/spring-boot-starter.md
@@ -229,6 +229,20 @@ Available properties:
       <td><code>true</code></td>
     </tr>
     <tr>
+      <td><code>use-create-time</code></td>
+      <td>
+        Specifies whether tasks should be fetched based on their create time in descending order. Use this property in disjunction with order-by-create-time property or a SpringExternalTaskClientException will be thrown.
+      </td>
+      <td><code>false</code></td>
+    </tr>
+    <tr>
+      <td><code>order-by-create-time</code></td>
+      <td>
+        Specifies whether tasks should be fetched based on their createTime with the given configured order. It can be either "asc" or "desc". Use this property in disjunction with order-by-create-time property or a SpringExternalTaskClientException will be thrown.
+      </td>
+      <td><code>null</code></td>
+    </tr>
+    <tr>
       <td><code>async-response-timeout</code></td>
       <td>
         Asynchronous response (long polling) is enabled if a timeout is given.


content/user-guide/ext-client/spring-boot-starter.md
```diff
@@ -228,6 +228,20 @@ Available properties:
       </td>
       <td><code>true</code></td>
     </tr>
+    <tr>
+      <td><code>use-create-time</code></td>
+      <td>
+        Use createTime when fetching and locking external tasks.
```
psavidis, line 234:
```suggestion
        Specifies whether tasks should be fetched based on their create time in descending order. Use this property in disjunction with order-by-create-time property or a SpringExternalTaskClientException will be thrown.
```

psavidis: ❌ According to the `ClientFactory` that [configures](https://github.com/camunda/camunda-bpm-platform/blob/5df05145727a26c7b207f3c031d03a9bbc1a5076/spring-boot-starter/starter-client/spring/src/main/java/org/camunda/bpm/client/spring/impl/client/ClientFactory.java#L109) the spring external task client, `use-create-time` property will configure the client to fetch tasks by their `createTime` in `DESC` order.

We can reflect that in the description and state also that this property has to be used in disjunction with the `order-by-create-time` property, otherwise `SpringExternalTaskClientException` will be thrown.

- `use-create-time` is a convenient flag that follows the pattern of `use-priority` for convention backwards-compatibility sake to our users which automatically orders descending
- `order-by-create-time` can be used to control explicitly the order

content/user-guide/ext-client/spring-boot-starter.md
```diff
@@ -228,6 +228,20 @@ Available properties:
...
+      </td>
+      <td><code>false</code></td>
+    </tr>
+    <tr>
+      <td><code>order-by-create-time</code></td>
+      <td>
+        Specifies whether tasks should be fetched based on their createTime with a configured order. It can be either "asc" or "desc".
```
psavidis, line 241:
```suggestion
        Specifies whether tasks should be fetched based on their createTime with the given configured order. It can be either "asc" or "desc". Use this property in disjunction with order-by-create-time property or a SpringExternalTaskClientException will be thrown.
```

psavidis (changes requested):
Thanks for adding the create time documentation to the spring boot section! I've
- added some reference to how the configuration of the two properties behaves (it's one or the other, both will throw an exception) and 
- added suggestions that explain the above

content/user-guide/ext-client/spring-boot-starter.md
joaquinfelici: @psavidis Very clear, thanks! Updated with your suggested changes.

psavidis (approved):

PR #1640, opened 2024-05-27
docs(update) Clarify support for Jakarta EL

venetrius:
related to: camunda/camunda-bpm-platform/issues/3992

PR #1648, opened 2024-06-13
docs(update) Clarify support for Jakarta EL (#1640)

venetrius:
related to: camunda/camunda-bpm-platform/issues/3992


venetrius:
After merge docs still runs locally. 
As per our new back port policy merging `cherry-pick` does not require review.

PR #1665, opened 2024-07-24
Update http-connector.md for  Camunda HTTP Connector enhancement for 4XX and 5XX errors -GIT Issue 4241

Nandanrshenoy:
added documentation updates for http-connector error handling enhancement git issue : https://github.com/camunda/camunda-bpm-platform/issues/4241

Change Description:
Current HTTP connector does not throw an error for 4XX and 5XX errors
The errors need to be explicitly handled by the modeler through an inline script.
Since this code of handling HTTP response is repeatable and reusable, it would be good to modify default behavior of http connector to throw a bpmnError in case of API call failures so that modeler can handle them as desired.
This has been made configurable through an engine setting by adding a new flag called throw-http-error. By default, this flag would be false, where modeler would have to explicitly handle such Errors. If marked true, then the feature that is being built to handle such errors should be enabled.

### DIFF
--- a/content/reference/connect/http-connector.md
+++ b/content/reference/connect/http-connector.md
@@ -109,12 +109,10 @@ HttpResponse response = http.createRequest()
   .execute();
 ```
 
-## Enabling HTTP Error Handling
+## Enabling HTTP Response Error Handling
 
-<strong>Note:</strong> HTTP Connector does not seamlessly handle 4XX and 5XX related response errors during HTTP call.
-Custom scripts within workflows have to be employed to handle these errors.To support handling of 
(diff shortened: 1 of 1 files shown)


content/reference/connect/http-connector.md
```diff
@@ -109,6 +109,21 @@ HttpResponse response = http.createRequest()
   .execute();
 ```
 
+## Enabling HTTP Error Handling
```
danielkelemen, line 112: 🔧

```suggestion
## Enabling HTTP Response Error Handling
```

danielkelemen (commented):
👍 Looks nice! I added two minor improvements.

danielkelemen (approved):

PR #1698, opened 2024-11-06
docs(mig-guide): allow query for hist pis by async activities

tasso94:
related to camunda/camunda-bpm-platform#4769


venetrius (approved):
👍 Looks good to me!

PR #1699, opened 2024-11-07
chore(update): add note about public repo config

PHWaechtler:
related to [#4744](https://github.com/camunda/camunda-bpm-platform/issues/4744)

Adds a note about configuring the public repo for accessing `camunda-license-check` required since 7.22.

### DIFF
--- a/content/update/ce-to-ee.md
+++ b/content/update/ce-to-ee.md
@@ -86,7 +86,7 @@ To officially begin migrating from Camunda 7 Community Edition to the Enterprise
         <id>Camunda-bpm-nexus-ee</id>
         <name>camunda-bpm-nexus</name>
         <url>
-        https://camunda.jfrog.io/artifactory/private/
+        https://artifacts.camunda.com/artifactory/private/
         </url>
       </repository>
     </repositories>
--- a/content/update/ce-to-ee.md
+++ b/content/update/ce-to-ee.md
@@ -72,9 +72,16 @@ To officially begin migrating from Camunda 7 Community Edition to the Enterprise
       <version>7.15.0-ee</version>
     </dependency>
     ```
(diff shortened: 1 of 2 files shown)


content/update/ce-to-ee.md
```diff
@@ -72,9 +72,16 @@ To officially begin migrating from Camunda 7 Community Edition to the Enterprise
...
+      <repository>
+        <id>camunda-bpm-nexus</id>
+        <name>camunda-bpm-nexus</name>
+        <url>
+        https://artifacts.camunda.com/artifactory/public/
+        </url>
+      </repository>
       <repository>
         <id>Camunda-bpm-nexus-ee</id>
         <name>camunda-bpm-nexus</name>
```
joaquinfelici, line 87: ❓ The public repo looks correct (consistent with our [maven coordinates](https://docs.camunda.org/get-started/apache-maven/#community-edition-1) for the public repo), but now that I look at it, the url for the private repo is different [from this one](https://docs.camunda.org/get-started/apache-maven/#enterprise-edition-1), do you think it's simply a redirection, or is it two different links?

joaquinfelici (commented):
Looks good! Just added one comment about something I didn't catch at first.

content/update/ce-to-ee.md
PHWaechtler: Hm, they both lead to jfrog for me so looks like a redirect, but I'll double check with infra just in case.

PHWaechtler: They are both the same, but we should not be mentioning the jfrog.io naming so I will adjust this. See [here](https://camunda.slack.com/archives/C5AHF1D8T/p1731064017049339?thread_ts=1731045840.505599&cid=C5AHF1D8T)

joaquinfelici: Thanks!

joaquinfelici (approved):
Looks good 👍

PR #1732, opened 2025-01-15
feat(docs): add queue capacity property for fetch and lock long polling

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4885


venetrius (approved):
👍

PR #1733, opened 2025-01-17
mention that oauth2 idp has no tenant concept

jonathanlukas:


danielkelemen (approved):
👍 Thanks for updating!

PR #1758, opened 2025-03-24
chore(activityView): explain cancel activity badge

PHWaechtler:
related to #4672

While checking the docs for where to add this I also noticed a typo for the call activity overlay that this fixes (the overlay is on the bottom not the top corner)


venetrius (approved):
👍 Looks good to me!

PR #1759, opened 2025-03-25
chore(docs): migrate modeler c7 only docs from c8

christinaausley:
Based on https://github.com/camunda/camunda-docs/pull/5312#issuecomment-2750702708.

Moving defining templates and HTTL back into C7 from the C8 docs.

Still to do:

- [x] Update tables? https://github.com/camunda/camunda-docs/issues/5380
- [x] I could not find a good place for the flags -- @barmac do you have thoughts?
- [x] I migrated over the HTTL doc as well -- @barmac do you know where you would like this placed?


barmac:
Thank you @christinaausley for your work on this.

I noticed that the images wouldn't load due to `hugo` specific way of handling them, so I fixed the paths. I'd like to move the "Defining templates" section to "Element templates", and bring back some kind of introduction. I can handle that beginning next week.

For the flags, we could add a page in Modeler section that will redirect the user to C8 docs. I would rather not have separate pages documenting some of the flags in C7 docs, and some in C8 docs. All of the flags (including the one on history time to live) should be documented in one place as they belong to the same application, even if it means that a C7-specific flag leaks to C8 docs. What do you think?

> I migrated over the HTTL doc as well

Thank you. I think this page may be placed in the Modeler section, and link back to the process engine docs for technical details.

barmac:
> So you would prefer to keep the C7-related flags in the C8 docs? I'm fine with this exception.

Correct!

> I'll move this in an incoming commit. Do you have a preference for where in the Modeler section? Perhaps Reference, unless you want this nested below a particular area in the user guide based on the developer journey.

I meant this section: https://docs.camunda.org/manual/latest/modeler/

barmac:
@christinaausley I've restructured the element templates part. Please have a look. From my perspective, it's ready to be merged.

barmac (approved):

christinaausley:
Really awesome work @barmac ⚡

PR #1775, opened 2025-04-07
fix(webapps): point to correct migrationError screenshot reference

venetrius:
related to: camunda/camunda-bpm-platform/issues/4830


yanavasileva (approved):
👍

PR #1781, opened 2025-04-22
fix(download): fix community download link to point to C7 page

HeleneW-dot:
related to [#5123](https://github.com/camunda/camunda-bpm-platform/issues/5123)


venetrius (approved):
👍 Looks good to me!1
I am not sure if the `bug` label is really necessary for this issue. Maybe `scope:documentation` is enough. 
You might also consider adding version labels to the issue, although since this is just a documentation change, it's not strictly required.

PR #1803, opened 2025-07-14
chore(docs): switch from quartz to spring cron expressions

yanavasileva:
https://github.com/camunda/camunda-bpm-platform/issues/4944


joaquinfelici (approved):
👍

PR #1816, opened 2025-08-26
docs(patch-migration): C7 Run 7.21.13 uses Spring Boot 3.5

tasso94:
related to camunda/camunda-bpm-platform#5367


tasso94:
We don't have a compatibility matrix for C7 Run. Just for the starter. We are not changing the starter.

PR #1832, opened 2025-10-09
fix(docs): fix start time event example to match behavior

joaquinfelici:
Related to https://github.com/camunda/camunda-bpm-platform/issues/5476


yanavasileva:
@joaquinfelici, I think this can be closed, right?

PR #1844, opened 2025-11-24
chore(workflow): remove experimental merge method detection option

tasso94:


HeleneW-dot (approved):
Lgtm, now the question is can we backport this? 😄

tasso94:
😄 I think we don't need it since workflows are always picked up from the default branch.

PR #1845, opened 2025-11-24
docs(wildfly): document downsized WildFly distribution

tasso94:
* Remove migration guides that left the extended support period.
* Improve patch level guide
  * Change ordering from newest to oldest
  * Make patch-level changes expandable

related to camunda/team-automation-platform#406


HeleneW-dot (approved):
Cant say anything about the css files but lgtm locally 👍

